PR #13, opened 2016-12-06
Added travis ci support for node 7

amilajack:


.travis.yml
```diff
@@ -1,5 +1,6 @@
 language: node_js
 node_js:
+  - 7
```
andywer: Since even node versions are the LTS releases: Replace `latest` by `7` and skip the `5` altogether?

amilajack: Good idea

amilajack:
Fixed

andywer:
Ahh, I meant "replace '7' with 'latest'". But nevermind, explicit versions are good ;)

Thanks for contributing!

PR #20, opened 2016-12-07
Add empty extension to the default extensions

eXon:
When trying to use a file including the extension, it will crash unless you have an empty extension. It is a pretty common Webpack configuration.

Before the fix this configuration was throwing an error: 

```js
const path = require('path');
const { createConfig, entryPoint, env, setOutput, webpack } = require('@webpack-blocks/webpack');
const babel = require('@webpack-blocks/babel6');

const config = createConfig([
  entryPoint('./src/index.js'),
  setOutput('./build/bundle.js'),
  babel(),
]);

module.exports = config;
```

**ERROR in multi main
Module not found: Error: Cannot resolve 'file' or 'directory' ./test.js in /private/var/git/thereactivestack/webpack-blocks-test/src
 @ multi main**

But it was working if I was using `./src/index` instead without the extension.

### DIFF
--- a/packages/webpack/__tests__/integration.test.js
+++ b/packages/webpack/__tests__/integration.test.js
@@ -76,7 +76,7 @@ test('complete webpack config creation', (t) => {
 
   t.is(webpackConfig.devtool, 'cheap-module-source-map')
 
-  t.deepEqual(webpackConfig.resolve.extensions.sort(), [ '.js', '.json', '.jsx' ])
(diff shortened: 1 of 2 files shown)


andywer:
Thank you very much @eXon! This is a webpack 2 change that slipped into the webpack 1 block... :-/

I will add an issue for end-to-end testing (actually running webpack on a sample project) to avoid this kind of regression in the future!

andywer:
Published as `@webpack-blocks/webpack v0.1.5` 👍

PR #21, opened 2016-12-08
Fix dev-server when using named entries

eXon:
When adding `devServer` on an entry like that: `entryPoint({ app: './src/index' })`, it would override the whole entry point by only the devServer.

This is fixing the way the entry is merged. 

Here is my whole webpack.config.js:

```js
const path = require('path');
const { createConfig, env, entryPoint, setOutput, addPlugins, setDevTool, webpack } = require('@webpack-blocks/webpack');
const babel = require('@webpack-blocks/babel6');
const cssModules = require('@webpack-blocks/css-modules');
const extractText = require('@webpack-blocks/extract-text');
const { StatsWriterPlugin } = require('webpack-stats-plugin');
const devServer = require('@webpack-blocks/dev-server');

const development = [
  setOutput({ filename: '[name].js' }),
  setDevTool('eval'),
  addPlugins([
    new webpack.DefinePlugin({ __DEV__: JSON.stringify(true) }),
    new webpack.NoErrorsPlugin(),
  ]),
  devServer('webpack-hot-middleware/client'),
  devServer.reactHot(),
];

const production = [
  setDevTool('source-map'),
  addPlugins([
    new webpack.DefinePlugin({ __DEV__: JSON.stringify(false) }),
    new webpack.optimize.DedupePlugin(),
    new webpack.optimize.CommonsChunkPlugin({
      name: 'vendor',
      minChunks(module) { return /node_modules/.test(module); }
    }),
    new webpack.optimize.UglifyJsPlugin({
      compress: { warnings: true },
      screwIe8: true
    }),
  ]),
  extractText('style.[contenthash].css'),
];

  entryPoint({ app: './src/index' }),
  setOutput({
    path: path.resolve('./build'),
    filename: '[name].[hash].js',
    publicPath: '/assets/'
  }),
  babel(),
  cssModules(),
  addPlugins([
    new StatsWriterPlugin({ fields: ['publicPath', 'assetsByChunkName'] }),
    new webpack.DefinePlugin({ 'process.env': JSON.stringify(process.env), }),
    new webpack.DefinePlugin({
      __CLIENT__: JSON.stringify(true),
      __SERVER__: JSON.stringify(false),
    }),
    new webpack.optimize.OccurrenceOrderPlugin(),
  ]),
  env('development', development),
  env('production', production),
]);
```

### DIFF
--- a/packages/core/src/__tests__/entryPoint.test.js
+++ b/packages/core/src/__tests__/entryPoint.test.js
@@ -1,0 +1,23 @@
+import test from 'ava'
+import { entryPoint } from '../index'
+
+test('entryPoint() should normalize string to object entry', (t) => {
+  let config = {}
+
+  config = entryPoint('./test.js')(null, config)
+  t.deepEqual(config.entry.main, { main: ['./test.js'] })
+})
+
+test('entryPoint() should normalize array to object entry', (t) => {
+  let config = {}
+
+  config = entryPoint(['./test.js'])(null, config)
+  t.deepEqual(config.entry.main, { main: ['./test.js'] })
+})
+
+test('entryPoint() should keep intact object entry', (t) => {
+  let config = {}
+
+  config = entryPoint({ main: './test.js' })(null, config)
+  t.deepEqual(config.entry.main, { main: ['./test.js'] })
+})
--- a/packages/dev-server/__tests__/devServer.test.js
+++ b/packages/dev-server/__tests__/devServer.test.js
@@ -1,0 +1,25 @@
+import test from 'ava'
+import devServer from '../index'
+
+test('devServer() add himself to every entries', (t) => {
+  let config = {
+    entry: {
+      main: ['./test.js'],
+      second: ['./second.js']
+    }
+  }
+
+  config = devServer('entry')(null, config)
+  t.deepEqual(config.entry, { main: ['entry'], second: ['entry'] })
+})
+
+test('devServer() use webpack/hot/only-dev-server as default', (t) => {
+  let config = {
+    entry: {
+      main: ['./test.js']
+    }
+  }
+
+  config = devServer()(null, config)
+  t.deepEqual(config.entry.main, ['webpack/hot/only-dev-server'])
+})
--- a/packages/dev-server/index.js
+++ b/packages/dev-server/index.js
@@ -40,39 +40,39 @@ function devServer (options, entry) {
   })
 }
 
-function addDevEntry(devServerEntry, config) {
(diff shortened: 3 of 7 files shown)


andywer:
Good point! But I see two issues here:

1. Generally, it is a tricky task to merge the entry points if you have got an entry points object like `{ app: './src/app.js', vendor: [ 'react', 'react-dom', ... ] }` and a string like `'./src/foo'`

Maybe entry points should always be normalized into an object... Forcing the user to pass an entry points object might be the most straight-forward solution here (`entry('./src/app')` => `entry({ app: './src/app' })`)

Making the `dev-server` pass an object as well is trivial and would solve the issue (or is there another issue I missed?).

2. The code in the PR seems much too complex. There is a better way. Maybe the object-only idea.

Thank you already for your efforts! :)

eXon:
The entry point `./src/foo` could be normalized to `{ main: './src/foo' }`. `main` is the default name of the chunk.

The main issue is webpack-merge doesn't merge the entry point correctly because it is not obvious and is a special case. You need to add the dev-server entry on every entry and you have to transform non-array entries to array entries.

I could make the code a little more straightforward if it was always saved as an object and the entries were always arrays. I'll try.

andywer:
I don't know if I got you completely right, but based on our last posts this might be what's necessary to get an elegant solution:

* Turn strings/arrays passed to `entryPoint()` into an object using `main` as default key
* Now we have got an entry point object and make sure all its values are an array

Btw: I don't think you need to add the dev-server entry to each entry. Adding it to the object using a new key should be fine. But we should double-check.

andywer:
@eXon Btw: I am really glad you brought that up. This entry-point issue can cause a lot of pain and resolving it early and in an elegant way will prevent dozens of issues later :)

eXon:
In the [webpack-hot-middleware docs(https://github.com/glenjamin/webpack-hot-middleware#use-with-multiple-entry-points-in-webpack) it is explicitely said you need to.

Let's say you have a home entry and a userProfile entry for two different pages. You'll need the dev-server on both if you want it to work. Not sure what would happen if you have two app on the same page though.

eXon:
I added the normalization to an object and each of them are within an array.

I also took the liberty to add a few unit tests. It's almost free with functional programming 👍

andywer:
Cool :) I think it's way better now!

I still have some minor changes in mind. Since I feel this is a pretty important we are working on here I will merge this into a local feature branch, apply some minor changes as well and open another PR for merge into `master`.

You can review my changes in the follow-up PR if you like, @eXon :)

PR #26, opened 2016-12-09
Normalize entry points, fix dev-server entry point merging

andywer:
This is the follow-up PR for #21.

It's basically the previous PR plus some minor code changes, a little more documentation, a few more tests and a version bump to `0.2.0` on `dev-server` and `webpack` blocks.

@eXon Feel free to provide feedback on my commits as well :)


packages/dev-server/README.md
```diff
@@ -28,6 +28,13 @@ module.exports = createConfig([
 ```
 
 
+## Pitfalls
+
+Make sure you use this `dev-server` block after your `entryPoint()` blocks. The dev server will extend all of them with the `'webpack/hot/only-dev-server'` entry point by default.
```
eXon: This is essentially false. Your dev-server can be after or before your main entry point. It has no impact what so ever. I used to put them at the beginning and it made no difference.

eXon:
I think this is great. Normalizing the entry should pretty much resolve any issue we might have with it right now.

packages/dev-server/README.md
andywer: Are you still able to order them randomly after we did the entry point normalization? Since we now take into account the existing entry points it should not work this way anymore, I think.

eXon: Now that it is normalized webpack-merge is taking care of merging everything correctly. It shouldn't be a problem if the dev-server is before or after the entry point.

andywer: That's right, the order is not important.

But I mean something else: `dev-server` adds itself to every `entry` key it finds in the config. If you define entry points (chunk name !== 'main') after `dev-server` then `dev-server` will not have added the dev-server-entry to this later entry's key.

eXon: Yes you have a point. It would be much better to use a post hook :)

andywer:
Published as `@webpack-blocks/dev-server v0.2.0` and `@webpack-blocks/webpack v0.2.0`

Issue #23, opened 2016-12-08
postcss webpack block


alextrastero:
Hiya,
I'm wondering here, the sass webpack block `index.js` defines the following:
`loaders: [ 'style-loader', 'css-loader', 'sass-loader' ]`
And the postcss webpack block is so:
`loaders: [ 'postcss-loader' ]`
But the postcss-loader specifies that you need `style-loader` and `css-loader`.
So my question is, how can I add a separate loader without using the "blocks" approach to make this work?

Only using the post-css webpack block I get the following error:
```js
var content = require("!!./../node_modules/css-loader/index.js!./App.css");
ERROR > if(typeof content === 'string') content = [[module.id, content, '']];
```
Which I think is because I don't have the style nor css loader installed.

Thanks in advance!

alextrastero:
My `webpack.config.js`
```JS
const { addPlugins, createConfig, entryPoint, env, setOutput, sourceMaps, webpack } = require('@webpack-blocks/webpack')
const babel = require('@webpack-blocks/babel6')
const postcss = require('@webpack-blocks/postcss')
const devServer = require('@webpack-blocks/dev-server')
const cssnext = require("postcss-cssnext")

const path = require('path')
const autoprefixer = require('autoprefixer')

module.exports = createConfig([
  entryPoint('./src'),
  setOutput({
    path: path.join(__dirname, 'dist'),
    filename: 'bundle.js',
    publicPath: '/dist/'
  }),
  addPlugins([
    new webpack.DefinePlugin({
      'process.env': JSON.stringify(process.env || 'development')
    })
  ]),
  babel(),
  postcss([
    cssnext(),
    autoprefixer({ browsers: ['last 2 versions'] })
  ]),
  env('development', [
    devServer(),
    sourceMaps(),
    devServer.reactHot(),
  ]),
])
```

andywer:
Hi @alextrastero.

The `loaders: [ 'style-loader', 'css-loader', 'sass-loader' ]` vs. `loaders: [ 'postcss-loader' ]` seems a little odd, but makes sense, because
* the postcss block is extending the `.css` loaders (which by default already got the `style-loader` and `css-loader`)
* while the sass block needs to set `style-loader` and `css-loader`, because there are no defaults set for `.scss|.sass`

I am stuck, though, concerning the rest of your question. Do you want to use `postcss` on the `sass` files and it does not work?

alextrastero:
[...]

PR #28, opened 2016-12-09
Fix PostCSS loader order issue + support postcss-loader v1.2.0

andywer:


andywer:
Published. Hopefully fixes #23.

PR #30, opened 2016-12-10
Bugfix: env() merges too much

### LINKED ISSUE #29: `env()` should pass `config` to setters, but not merge it already
`env()` returns a setter function (basically a block). This function receives the merge-in-progress `config` like any other block and that's fine.

But: It calls `invokeConfigSetters()` which is not only merging the `env()`'s blocks, but also the `config` passed by `createConfig()`, thus the result of `env()` has already merged all previous config into the webpack config snippet it returns.

Result: Duplicate stuff in the snippet created by `env()`.
Seems not be very critical, but it is a bug and might lead to unexpected behavior.

andywer:
Fixes #29.

### DIFF
--- a/packages/core/CHANGELOG.md
+++ b/packages/core/CHANGELOG.md
@@ -1,0 +1,13 @@
+# @webpack-blocks/core - Changelog
+
+## 0.1.2
+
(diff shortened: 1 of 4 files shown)


andywer:
Published as `@webpack-blocks/core v0.1.2`

PR #31, opened 2016-12-10
fixed cannot run test-app script

nndung179:
**Problem:**
+ I pulled webpack-block to my local machine. 
+ cd to test-app
+ After run `npm i`, I could not run `npm start`, it gave me message below

```
(function (exports, require, module, __filename, __dirname) { const {
                                                                    ^

SyntaxError: Unexpected token {
    at exports.runInThisContext (vm.js:53:16)
    at Module._compile (module.js:404:25)
    at Object.Module._extensions..js (module.js:432:10)
    at Module.load (module.js:356:32)
    at Function.Module._load (module.js:313:12)
    at Module.require (module.js:366:17)
    at require (module.js:385:17)
```

**Solution**
+ I have figured it out that the config of webpack in `test-app` uses syntax of JS ES6. So that, I decided to change file name of webpack config from: **webpack.config.js** to **webpack.config.babel.js** and change scripts in package.json. Now it worked well


andywer:
Hey @nndung179 for your PR! Good catch.

But don't we need `babel-register`, too now?

nndung179:
I discovered that. When you run `npm i`. Look in insider **node_modules**, it was already there.

andywer:
Ok, I see that `babel-core` requires `babel-register`, so it will work with node 5+.

The problem is: I tried it locally (changing the file name to `webpack.config.babel.js`) and it still fails with the same error... Did you do something else?

The other way to fix it would be to just get rid of the object destructuring assignment in the first three lines.

andywer:
Or... another quick-fix, since it's just the test app:

We could just add `{ "engines": { "node": ">= 6.0" } }` to the `package.json`... 🤔

nndung179:
I think I will stick with the first one: replace **destructuring** to **object.attribute**

andywer:
@nndung179 Nevermind. My mistake that it didn't work for me locally... 🙈😉

Let's take your fix and just add `{ "engines": { "node": ">= 5.0" } }` to the `package.json`, since it will not work with node 4.x (because we need "flat" node_modules, which requires npm 3.x which ships with node 5+) :)

test-app/package.json
```diff
@@ -1,7 +1,7 @@
 {
   "scripts": {
-    "build": "NODE_ENV=production webpack --config webpack.config.js",
-    "start": "NODE_ENV=development webpack-dev-server --config webpack.config.js"
+    "build": "NODE_ENV=production webpack --config webpack.config.babel.js",
+    "start": "NODE_ENV=development webpack-dev-server --config webpack.config.babel.js"
   },
   "devDependencies": {
```
andywer: Please add `{ "engines": { "node": ">= 5.0" } }` and this is ready for merging :)

nndung179: ok right a way

andywer:
👍👍

I just re-ordered stuff in the package.json as I recognized I can actually commit changes from within this PR site :)

nndung179:
I just known too :). Thank mate!

Issue #16, opened 2016-12-07
Discussion: Support for DLL


nndung179:
DLL plugins helps webpack not to loop through dependencies when webpack rebuild itself. Are you guys supporting this features?

andywer:
Hey @nndung179. Thanks for pointing that out.

There is no direct support yet. It would be no problem to use https://www.npmjs.com/package/@webpack-blocks/webpack#addpluginsplugins-webpackplugin, though. (Used like `createConfig(..., addPlugins([ new webpack.DllPlugin(...) ]))`)

**Seems like a low hanging fruit for someone eager to implement his first webpack block... 🤔😉**
(Doesn't have to be you, but could be you)

andywer:
PS: Or maybe not a new block, but rather another helper method for the `webpack` block.

nndung179:
Hey @andywer I think it should be in core of `webpack` block. It is kind of important in development env.

nndung179:
@andywer Could I handle this block?

andywer:
@nndung179 Sure! Just open a PR :)

You just have to add another function to [packages/webpack/index.js](https://github.com/andywer/webpack-blocks/blob/master/packages/webpack/index.js) and export it, I think.

Maybe something like
```js
function dll (<someOptions>) {
  return addPlugins(new webpack.DllPlugin(...))
}
```
(so it would basically be just a wrapper around `addPlugins()`)

PR #32, opened 2016-12-10
completed add dll plugins

nndung179:
I added **setDllPlugin** in webpack packages. If it is ok, I will update document

### DIFF
--- a/packages/webpack/index.js
+++ b/packages/webpack/index.js
@@ -20,6 +20,7 @@ exports.setContext = setContext
 exports.setDevTool = setDevTool
 exports.setOutput = setOutput
 exports.sourceMaps = sourceMaps
+exports.setDllPlugin = setDllPlugin
 
 /**
  * Wraps @webpack-blocks/core's `createConfig` to provide some sane default
@@ -139,4 +140,13 @@ function setOutput (output) {
  */
 function sourceMaps (devtool) {
   return setDevTool(devtool || 'cheap-module-source-map')
+}
+
+/**
+ * @see http://webpack.github.io/docs/list-of-plugins.html#dllplugin
+ */
+function setDllPlugin (options) {
+  return addPlugins([
+    new webpack.DllPlugin(options)
+  ])
 }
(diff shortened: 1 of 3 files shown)


eXon:
It's great to have DLLs but it's better when you can reference them ;)

You probably should add a core block for `DllReferencePlugin`. Looking forward to using it 👍

andywer:
Hey guys! Interesting to see that the DLL functionality is such a hot feature.

But the proposed `setDllPlugin` function doesn't make it any easier to use it, I think. Is `setDllPlugin(x)` so much more readable than `addPlugins([ new webpack.DllPlugin(x) ])`?

(Feel free to share your opinion anytime if you feel differently about this! 😉)

Using the test app to demonstrate how the DllPlugin can be used seems like a good idea on first glance, but it also makes it more complicated. There is another webpack config file, another `script` in the `package.json` and so it doesn't feel as light-weight anymore.

I would suggest we provide a separate test app for the DLL functionality.

andywer:
Or even better: I don't know if you got time for this, but it might be really cool to write a short blog post / tutorial (maybe on medium.com) about why and how to use the DllPlugin and provide an easy to use and easy to understand configuration like this one here.

It might even live in a separate examples repository and we just link there.

Or maybe we can somehow make it simpler to use the DLL feature...

andywer:
Just recognized this is the follow-up of #16.

And I proposed exactly this, now I am not satisfied. Sorry man! 🙈😅

I guess I expected the util function to simplify the DllPlugin's options some more...

Sorry for the inconvenience! I appreciate your efforts.

nndung179:
I will try another way. Just figured it new way to do it. You are totally right about how easy to use this feature :)

nndung179:
Hi @andywer. I am making a function allow return dll configs. And I cannot think about how to run it without npm script. Any suggestion?

andywer:
Hey @nndung179!
Anything new here or can I close the PR for now?

nndung179:
I think PR should br closed. I will get back another time

andywer:
Ok 👍

And thx for the quick reply :)

Issue #25, opened 2016-12-09
Block internals: Pre- and post-processing hooks


andywer:
As @eXon pointed out in #17 it might be beneficial to enable webpack blocks to define post-processing functions. That could allow you to define constants multiple times in multiple blocks, but eventually get a single `webpack.DefinePlugin` instance.

Another use case is the `dev-server` block which is currently being changed (#21). It adds a dev-server entry point to every chunk in the `entry` object. Right now this means that the `dev-server` is used after setting all your entry points. A post-processing step defined by `dev-server` would allow you to use your `entry` and `dev-server` blocks in random order.

A pre-processing might be desirable to add custom mime types before any the blocks are run.

Open for discussion :)

andywer:
Another feature directly related to this: Have a context object in the webpack config object or independent from it to store metadata to share between blocks.

Another use case: `sourceMap()` currently enables source maps on JS code only. With a pre-process step it could set a flag on the context object to indicate to other blocks to enable source maps on the `css-loader` as well, for example.

eXon:
It certainly need a context object that every block can access. I happen to think the fileTypes argument should disappear.

The context could contain the fileTypes and more and it would be shared between every blocks.

The big question here in my opinion is how do we hook those block to pre/post processing? Are they doing it as soon as you import the Node.js module? Are we hooking when you hook a block for the first time using a unique key. Are we hooking using an EventEmitter within the block before returning the function?

I guess we'll have to playaround to see what works the best.

andywer:
Published in version 0.3.0.

Issue #24, opened 2016-12-08
Proper preset support


andywer:
In fact webpack-blocks already makes it quite easy to group blocks. But it would be more convenient if [@webpack-blocks/core](https://github.com/andywer/webpack-blocks/blob/master/packages/core/src/index.js) would export another function that works like `env()`, but without checking the environment.

## How it could look like

```js
const { group } = require('@webpack-blocks/core')
const babel = require('@webpack-blocks/babel6')
const postcss = require('@webpack-blocks/postcss')

const defaultPostcssPlugins = [ /**/ ]

module.exports = function myPreset ({ addPostcssPlugins }) {
  const postcssPlugins = defaultPostcssPlugins.concat(addPostcssPlugins || [])

  return group([
    babel(),
    postcss(postcssPlugins)
  ])
}
```

## How it could be used

```js
// webpack.config.js
const { createConfig, entryPoint, setOutput } = require('@webpack-blocks/webpack')
const myPreset = require('my-preset')

module.exports = createConfig([
  myPreset(),
  entryPoint('./src/app'),
  setOutput('./build/bundle.js')
])
```

eXon:
I think this would be very helpful.

It would be great to be able to write (same for `createConfig` actually): 

```js
groupBlocks(
  babel(),
  postcss(postcssPlugins)
)
```

andywer:
@eXon Hmm. I thought about that too when I started designing the API. But is there a real benefit?

Right now `createConfig()` takes just one parameter which is the blocks array. But at least there is an option to add a second one if there is a need to do so in future. If you pass every block as a single parameter this door will be shut forever.

I also think that it's more complicated to provide working type information (jsdoc, or in future maybe flow-type or so) for a function with an unlimited number of params.

andywer:
Note: Just renamed `groupBlocks` function => `group` (in the example at the top). It's clean, it's short and yet easy to understand :)

andreiglingeanu:
I'm having a question:

You guys are thinking about a way to `group` blocks. How do I override/disable/add something onto a `group`? 

Am I able to do something like this and expect that the `setOutput` block would be overided?

```javascript
let group1 = group(
  entryPoint('./src/app'),
  setOutput('./build/bundle.js')
);

let group2 = group(
  group1,
  setOutput('./another-build-dir/bundle.js')
);
```

[...]

Issue #17, opened 2016-12-07
Add a nice way to define constants


eXon:
It would be great to have at the core level of webpack-blocks a way to define constants within the code. Right now, we could do that using the DefinePlugin but I think it deserves a nicer way.

Here is my proposal:

```js
const { createConfig, defineConstant } = require('@webpack-blocks/webpack')

module.exports = createConfig([
  defineConstant('process.env.NODE_ENV', process.env.NODE_ENV),
  defineConstant('__API__URL', process.env.API_URL),
  // ...
])
```

And it would be transformed by:

```js
new webpack.DefinePlugin({
  'process.env.NODE_ENV': JSON.stringify(process.env.NODE_ENV),
  '__API_URL__', JSON.stringify(process.env.API_URL)
})
```

We might want to stack them instead of doing it with a single plugin. I don't know if it will be possible the way things are built with webpack-blocks.

Or maybe it could be a defineConstants that takes more than a single one.

What do you think? I can make a PR if you want to experiment with it.

andywer:
I also thought about the DefinePlugin, usability and edge cases, but without a clear answer so far.

Exporting a `defineConstant` helper function is basically no problem, but I am not sure if it really adds any value (except for having an easier-to-understand name).

What I am wondering right now is if you could have multiple instances of the DefinePlugin in your config or if that crashes or introduces noticeable performance penalties.

andywer:
@eXon When you wrote "want to stack them" and "don't know if it will be possible the way things are built with webpack-blocks" I guess you mean if it's possible to have multiple `defineConstant()` but get a single `new webpack.DefinePlugin` in the end?

eXon:
It is what I meant by stacking them. You can use multiple `DefinePlugin` but I don't know what's the performance impact. It is probably faster to have a single one.

Maybe the plugin could be added only when you call `createConfg` (haven't checked the implementation details yet).

That way, it is easier to override a value when you share configs between a client and server webpack config or if you want to share a common config between multiple projects. 

This is mainly syntax sugar but there is a few benefits when sharing configs.

andywer:
True. Being able to have multiple `defineConstants()` might be a real benefit in larger configs.

It is not achievable very easily right now without increasing code complexity under webpack-block's hood, I think. But there should be some way and it should be easy and elegant. Got to think about that...

eXon:
@andywer Here is an idea how it could be achieved without adding too much complexity:

[...]

PR #33, opened 2016-12-10
Proposal for pre/post hooks & context object & group()

eXon:
This is what I am proposing we could do for pre/post hooks. You can take a look at the dev-server for an example.

We could hook ourself before processing the blocks so that we can change the fileTypes or activate the sourceMaps before createConfig is called.

We can also hook ourself at the end so we can do something once we have the whole config (like the dev-server).

I also changed the `fileTypes` parameter by `context`. It is a big breaking change but I think it's for the best.

Let me know what you think! (the doc is still missing for now)


eXon:
I'm not sure about the additional config I added to `invokeConfigSetters`. It looks like I cannot use `baseConfig` to chain the calls so I added `initialConfig`. Not sure it's the best solution.

eXon:
It looks like there is a weird bug when merging the entry points. Both entry points are merged as `['./test.js', 'entry]` instead of there respective entry points. I have no clue why.

andywer:
Hey @eXon. Thanks for your case study on the pre/post-processing hooks feature.

I noticed some more issues about it though:
- Branch seems to be based on `improve/normalize-entry-points` and wants to be merged back into it. This is not so good. Feature branches should fork from `master`/`develop` and be merged back, so we don't get a mess there :)
- The global hooks arrays in the `core` block: Now the blocks would all need to require the `core` block and it will break as soon as there are multiple `@webpack-blocks/core` in `node_modules` (may happen if different blocks require different versions of `core`)

I like the idea of the `context` object and passing it as first parameter to the blocks, though 👍

andywer:
My suggestion for the pre/post-hooks:

Implementation in a block:
```js
function myBlock () {
  const setConfig = () => ({
    entry: './foo'
  })
  return Object.assign(setConfig, {
    pre (context) {},
    post (context, config) {}
  })
}
```

So it still returns a function as it always did, but we can also add `pre`/`post` properties to the function (since a function also is an object).

andywer:
Proposal: I am going to merge the `normalize-entry-points` soon, that's gonna be version `0.2.0`. We might bundle the `context` object and `pre`/`post` hooks feature together to a version `0.3.0`.

Btw: This is a PR for #25. Please mention the issue when opening a PR related to an issue 😉

eXon:
I'll change the PR target to master as soon as the branch is merged. I needed the change for testing with the dev-server but I didn't wanted to include those changes when looking at the diff.

I though about the same solution but what I don't like about it is we don't want to include the pre/post function more than once even though we might have a block multiple times.

The solution to this might be to switch the dependency to @webpack-blocks/core to a peerDependency. This would make sure even though you might have different versions on different blocks they would all use the same.

If you read about [peer dependencies](https://nodejs.org/en/blog/npm/peer-dependencies/), you'll quickly realized it is exactly for this kind of use case. The blocks are plugins and we want to make sure we include only once `@webpack-blocks/core` and respect a version requirement.

In my opinion it makes things a lot simplier. What do you think?

andywer:
The need to require a webpack-blocks package from within the blocks seems like an anti-pattern to me. And calling those functions outside of the block function as well.

We can make it a peer-dependency, but it is still not elegant and we introduce a footgun there. It will complicate the dependencies and as soon as there are multiple versions of the same package it will fall apart (a single `npm link` during development might already achieve that).

But: The objection that we might call the `pre`/`post` handlers multiple times is a good concern 👍👍

Got to find an elegant solution for that, the rest should be easy...

eXon:
@andywer We might be able to detect the pre/post hook is already there by comparing the function with toString() (not a big fan but would be effective)

It would be much simplier to write a pre/post hook and it would only run them if you used the block at least once. I'll try what it looks like.

eXon:
Here is an example of a possible post hook:

```js
function devServer (options, entry) {
  if (options && (typeof options === 'string' || Array.isArray(options))) {
    entry = options
    options = {}
  } else {
    entry = entry || 'webpack/hot/only-dev-server'
  }

  entry = Array.isArray(entry) ? entry : [ entry ]

  const devServerBlock = (context) => {
    context.devServer = true
    return {}
  }

  devServerBlock.post = (context, config) => {
    if (!context.devServer) {
      return {}
    }

    return {
      devServer: Object.assign({
        hot: true,
        historyApiFallback: true,
        inline: true
      }, options),
      entry: addDevEntryToAll(config.entry || {}, entry),
      plugins: [
        new webpack.HotModuleReplacementPlugin()
      ]
    }
  }

  return devServerBlock
}
```

The hook should be executed only once because it compares the function as a string. We can't use the reference because each of them will be different.

eXon:
If we choose to get this way, we'll need to make sure when we do #24 that we also set the subBlocks while merging them just like I did within `env()`

andywer:
@eXon Sorry for delayed response, but you know how it is in the time before x-mas... 🙈

Yeah, that should do the trick. I guess you don't like the `return Object.assign(() => ({}), { pre, post })` syntax? ^^ Do you think it's harder to read?

There might be one alternative to `function.toString()`: We could also define the `pre`/`post` methods outside the block function, so the references to the hook functions are constant. But it might be hard to track down the error if you create the `pre`/`post` functions dynamically.

eXon:
Just changed what you proposed and I think you'll like it better. In that example I keep no anonymous function for the sake of clarity.

```js
function devServer (options, entry) {
  if (options && (typeof options === 'string' || Array.isArray(options))) {
    entry = options
    options = {}
  } else {
    entry = entry || 'webpack/hot/only-dev-server'
  }

  entry = Array.isArray(entry) ? entry : [ entry ]

  const devServerBlock = (context) => {
    context.devServer = { entry }

    return {
      devServer: Object.assign({
        hot: true,
        historyApiFallback: true,
        inline: true
      }, options)
    }
  }

  return Object.assign(devServerBlock, { post: devServerPost })
}

function devServerPost (context, config) {
  if (!context.devServer) {
    return {}
  }

  return {
    entry: addDevEntryToAll(config.entry || {}, context.devServer.entry),
    plugins: [
      new webpack.HotModuleReplacementPlugin()
    ]
  }
}
```

andywer:
👍

Just one suggestion: `devServerPost` as a name feels not ideal. How about `postConfig`?
I think the function name does not need to contain `devServer`, since there is only dev-server stuff in the file :)

eXon:
I think it makes sense. 

If we can agree on that, I'm not sure where this should be documenting. This looks like an advanced feature when creating a block so I'm not sure if this should be within the main readme.

andywer:
@eXon I guess that's the right moment to turn the "How does a block look like from the inside?" paragraph of the README into a new separate markdown file: "How to write a webpack block" :)

I will have more time tomorrow and the following days. Gonna look into that 😉

eXon:
I just changed the target to master. I'll let you start the new docs folder.

Do you want to include the doc within this PR?

andywer:
Yes, I think at least some basic documentation should be part of this PR. Will look into it this weekend!

The PR is approaching its merge... :)

andywer:
Btw: Just realized that the probably most important part of integrating this PR will be the context-object breaking change.

It might be good to increase the version number of **all** blocks to `0.3.0` and put a big note in the top part of the README that you either have to upgrade all blocks to `>= 0.3.0` or use all blocks in a version `< 0.3.0`.

andywer:
@exon I still have some smaller code change proposals (I think we don't need the `subSetters`). Do you mind if I just push to your feature branch? (I think GitHub allows me to, but not completely sure)

eXon:
@andywer I'm not sure how we can get rid of subSetters. It keep track of the setters when you use `env()` or potentially latter `groups()`. Otherwise, there is no way to know what are the pre/post hooks within those grouped blocks.

andywer:
@eXon Actually it should be quite easy. Right now (before the PR) `group(a, b) = combineSetters(a, b)`.

Right now in the PR it is `group(a, b) = Object.assign(combineSetters(a, b), { subSetters: [a, b] })`

But this means we now have two types of blocks. Blocks with `subSetters` and blocks without. And since uniformity is key to composability, I propose:

`group(a, b) = Object.assign(combineSetters(a, b), { pre: combineHooks(a.pre, b.pre), post: combineHooks(a.post, b.post) })`

(I know, there is no `group()` yet, but it is just the 90% functionality of `env()` we are talking about here :) )

I think it's easy to see if I just adapt the code.

eXon:
I added you as a collaborator on my fork so I guess you have access to change it now

eXon:
@andywer I though about the same thing at first but it becomes a lot harder to remove duplicate hooks that way.

andywer:
@eXon Damn it. You are right.

But... I think I know another approach that might do the trick: Let's allow `pre` and `post` to be an array, so `group(a, b) = Object.assign(combineSetters(a, b), { pre: [ a.pre, b.pre ], post: [ a.post, b.post ] })` (slightly simplified).

It is slightly unelegant, because `pre`/`post` may now be either a function or an array of functions. But... There is no need for a new property, so the block data model can stay small and concise.

eXon:
@andywer I think I found a pretty elegant solution with your idea. Even more simplier than what we have and easily pluggable on the future `groupBlocks()` function :)

andywer:
@eXon 👍

I am right in the middle of implementing the same thing (incl. the unit tests to make sure it does what it's supposed to do).

I guess I will finish my work and then try to take your's + my tests.

eXon:
I was hoping you wasn't doing it at the same time xD My code is lacking tests for sure.

andywer:
Updated the feature branch / PR. And the headline 😅

Added unit tests and changed the code of `core/src/index.js` a bit. *Feel free to review what I did!*
Also implemented the `group()` method for creating presets, since it's < 10 lines of code and we now got everything we need to implement it.

andywer:
Btw: `config` is not passed to the `pre` hooks anymore, since the pre hooks run before configuration stuff and cannot change it either 😉 So we enforce pre hooks to really just modify the context.

andywer:
Looking at `test: context.fileTypes('text/css'),` and wondering if it should rather be `test: context.fileType('text/css')` (fileType, without the 's') 🤔

Then it would also be `context.fileType.add('foo/bar', /something/)`. But that seems ok :)

andywer:
@eXon Let's decide on `context.fileTypes` vs. `context.fileType` and merge that beast of a PR then.

Subsequent steps:
- Open PR for new / updated documentation
- Implement `defineConstants` (see #17)
- Version bump: 0.3 (a.k.a. the christmas release 🎄)
- (Port changes to webpack2 branch)

eXon:
@andywer I think plurial should be used if there is a good reason behind only. `context.fileType()` is slightly better.

andywer:
@eXon Alright. Will look into that tomorrow morning. Thinking about doing that in a separate PR after merging this one :)

We are approaching the christmas release! 🎉

andreiglingeanu:
Sounds great!

PR #35, opened 2016-12-11
Bugfix: Fix extract-text block

andywer:
Didn't work before when there were more than 2 webpack loaders assigned to a file type.

Also added another parameter, so you can now apply the ExtractTextPlugin on other file types as CSS (like SASS, for instance) as well.


andywer:
Published as `@webpack-blocks/extract-text v0.1.1`

Issue #34, opened 2016-12-11
webpack-merge or not?


andywer:
Right now webpack-blocks is using [webpack-merge](https://github.com/survivejs/webpack-merge) to merge the blocks' config snippets to a complete config object.

There are some problems with webpack-merge, though:
- It is designed to concatenate/extend config objects/properties/arrays, but it is a problem to clear an object that's part of the webpack config that has been set before.
- An array might be cleared by passing an empty array to the merge function. But this causes another issue: You wouldn't expect that merging an empty array clears all existing items.

The webpack-merge makers are already discussing the issue (survivejs/webpack-merge#45), but I am not a big fan of the proposed approach.

Possible solution: Use https://github.com/kolodny/immutability-helper instead of webpack-merge altogether.

Open for discussion :)

eXon:
Is webpack-merge doing anything special webpack related? If not that much, we could simply use [lodash.merge](https://lodash.com/docs/4.17.2#merge). There is even a package for just that function called [lodash.merge](https://www.npmjs.com/package/lodash.merge).

It's what I was using for Webpacker and it is working pretty well. Instead of doing `merge.smart(config1, config2)` you would do `merge({}, config1, config2)`. Maybe it's worth a try :)

AFAIK merge.smart is doing the same thing + merging loaders if they have the same regex. Not sure it is much need and would probably easy to replicate.

andywer:
@eXon Yeah, `merge.smart()` does merge the loaders, the entry pointers and maybe some more stuff.

The problem about `lodash.merge` as I think is that you will again run into the issues of clearing an array or object that has been set before by applying a diff. Or just doing more complicated changes based on existing props.

With the immutability helpers you could do this:
`{ plugins: { $push: [ ... ] } }`  (Add another plugin)
`{ module: { loaders: { $apply: someFunction } } }`  (`someFunction = (oldLoaders) => newLoaders`)

eXon:
I think we will need to normalize configs just like the entry point to make sure everything is merged correctly. There is no way around it.

I've been using immutable.js for some time so I understand there is some benefits to this.

andywer:
I think using the immutability helpers would give us much more power and at the same time make the webpack config merging details more explicit. There are two ways I see right now:

### The soft way

Having a simple heuristic to determine if a webpack config snippet is supposed to be handled by the immutability helper (by looking for `$`-prefixed command props). If not webpack-merge is used as before.

Pro:
[...]

PR #41, opened 2016-12-20
v0.3 docs

andywer:
Updated the README, added an update note for the soon-to-be-released v0.3 (a.k.a. christmas release 🎄) and added a whole new documentation file on how to create blocks.

The actual features have been introduced by #33 (thanks @eXon!). You can preview the changes of this PR [here](https://github.com/andywer/webpack-blocks/tree/feature/v0.3-docs).

Feedback of any kind is very welcome 😉

Poking @eXon @andreiglingeanu

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,9 +13,9 @@ Missing anything? Write your own and share them!
 
 <br />
 
-## v0.3 Update note
+## v0.3 Update
 
-**Version 0.3.x of webpack-blocks contains a bunch of new features, but also with a major breaking change under the hood.**
+**Version 0.3.x of webpack-blocks comes with a bunch of new features, but also with a major breaking change under the hood.**
 
 *Every* block has been updated, but you have to make sure **all webpack-blocks packages you use are version >= 0.3.0** or otherwise **all packages are < 0.3.0**.
 
--- a/README.md
+++ b/README.md
@@ -8,15 +8,17 @@ Missing anything? Write your own and share them!
 >
 >[Dan Abramov](https://github.com/gaearon) via [twitter](https://twitter.com/dan_abramov/status/806249934399881216) (Co-author of Redux, Create React App and React Hot Loader)
 
+<br />
 
 ## v0.3 Update note
 
-**v0.3 of webpack-blocks ships with a bunch of new features, but also with a major breaking change under the hood.**
+**Version 0.3.x of webpack-blocks contains a bunch of new features, but also with a major breaking change under the hood.**
 
 *Every* block has been updated, but you have to make sure **all webpack-blocks packages you use are version >= 0.3.0** or otherwise **all packages are < 0.3.0**.
 
 On the upside we don't consider it an early release anymore and the next major release you see might very well be `v1.0` 👌
 
+<br />
 
 ## Installation
 
@@ -30,7 +32,7 @@ npm install --save-dev @webpack-blocks/webpack @webpack-blocks/babel6 ...
 Create a development config with Babel support, dev server, HMR and PostCSS autoprefixer:
 
 ```js
-const { createConfig, env, entryPoint, setOutput, sourceMaps } = require('@webpack-blocks/webpack')
+const { createConfig, defineConstants, env, entryPoint, setOutput, sourceMaps } = require('@webpack-blocks/webpack')
 const babel = require('@webpack-blocks/babel6')
 const devServer = require('@webpack-blocks/dev-server')
 const postcss = require('@webpack-blocks/postcss')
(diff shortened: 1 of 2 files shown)


eXon:
This is a great starting point. What's look weird to me is it is a mix of an API docs and a guide docs. Maybe we should split it between API.md and multiple small guides with clear and precise intend? By doing so it will be easier to generate a doc with GitBook or something like that.

andywer:
Do you mean the README.md, BLOCK-CREATION.md or both?

Splitting API docs and guides seems like a good idea, but I think the README has got to cover both to some degree. Unless we keep the README minimalistic and just link to the other docs... 🤔

andywer:
Added a quote of Dan Abramov to the readme. What do you think, @eXon?

andreiglingeanu:
Looks great!

I like this way of composing configs a lot. Can you tell me, in two words, how the block specification is coupled with webpack? Will I be able to use same API for composing other types of configs?

As far as I know, the return value of blocks are merged with `webpack-merge` (and you are still in doubt if that's the right way to go), right?

andywer:
@andreiglingeanu Actually the core API isn't coupled to webpack at all, besides `webpack-merge`. And yes, I am still not sure if the smart merge of webpack-merge is the best way to go (#34).

Basically the webpack-blocks approach of - let's call it - "functional configuration" could be applied to pretty much everything. I already use it for another use case as well ([flux-capacitor-boot](https://github.com/flux-capacitor/flux-capacitor/tree/master/packages/flux-capacitor-boot)).

For composing large and complicated data structures (like the webpack config) which you cannot make much simpler I would use that approach any time. If you are actually writing your own API and think about applying it there it might be worth considering rather functional code design patterns for your API itself, like Observables (👉 keeping it functional and complexity low from the beginning, instead of fixing it later using something webpack-blockish).

Why are you asking? 😉

eXon:
@andywer I mean the README.md. I feel like it focus too much on how things are working rather than answering the question from a user experience.

I like how [react-router](https://github.com/reacttraining/react-router) has a docs folder focusing on one thing you might be trying to achieve, an examples folder and a very light readme just to get things started.

[Redux](https://github.com/reactjs/redux) is also doing the same thing and it generates a great Website using GitBook.

Maybe this should be in a separate issue, it might be out of scope for just documenting v0.3.0

andreiglingeanu:
@andywer I have composable task builder for `gulp`, which includes handling multiple webpack entries (and some [dynamic](https://github.com/Creative-Themes/build-process/blob/f9aae907ab5b6efa97dd590ea93b3b2e0f9b1fa5/webpack.config.js#L20-L76) ones). I use in multiple internal projects of mine and I really like to keep it flexible enough to be able to handle complex situations. I'd also like to be able to quickly bootstrap a new project and get it running in no time.

This builder basically is just a function that receives `gulp` as an argument and assigns tasks to it (`gulp` itself is just a proxied [`Orchestrator`](https://github.com/robrich/orchestrator#1-get-a-reference) instance). The way I attach those tasks is configurable and current configuration API is a mess. That's why I was looking for possible ways of enforcing a structure over it. I'll maybe introduce some basic building blocks like - `webpack`, `sass` and a couple of other regular build tasks I use frequently.


Damn, I'll maybe switch away from handling sass with `gulp` and just proceed with ExtractText when I need files to be saved separately from the `.js` bundle. But I didn't liked the idea to keep one more server running (webpack HMR) just to be able to receive live CSS updates in browser - I don't use HMR for my JS (or I didn't bothered to configure it properly :)).

Thank you for pointing out the fact that you are using this architecture approach (yeah, let's name it _functional configuration_) not just for `webpack-blocks`. I'll take a more closer look at it.

> keeping it functional and complexity low from the beginning, instead of fixing it later using something webpack-blockish

That's a very good point but, sadly, it is too late :(

andywer:
@eXon Pretty good point... Now that we have got presets we might also just provide a preset shipping basic pre-configured babel and dev-server config, plus setting `process.env.NODE_ENV`.

So we might end up with a good part of Create React App's features, but in a composable way.

But my personal goal for this week is shipping `v0.3` before christmas. We could probably release it today if we want, but I still try to get the webpack-2 stuff into `0.3` as well :)

On the other hand... If it already includes webpack-2, we could call it `v1.0` just as well 😅

andywer:
@eXon Sorry, I lost focus. You are right about "explaining how to do stuff with it" vs. "explaining how it does stuff". But I also agree on postponing it a little longer.

andywer:
@andreiglingeanu Not quite sure why you need gulp if use webpack, but I suppose you have got your reasons :)

> It really was simple from the beginning - but I started to add options in a hacky way so that I achieve my needs quickly.

Yeah, that's what usually happens if you go the data-driven configuration way. If it comforts you: I can easily imagine that the webpack creators experienced quite the same issue in webpack's early days just as well 😉

PR #42, opened 2016-12-20
Rename `context.fileTypes` => `context.fileType`

andywer:
Follow-up of #33. Need to update #41 as well.


eXon:
Luckily this is not a breaking change, switching to the context object is. We just had a variable name that had no impact before 👍

PR #43, opened 2016-12-20
Implement `defineConstants`

### LINKED ISSUE #17: Add a nice way to define constants

andywer:
Now that we have got `pre`/`post` hooks, `context` and `group()` it is easy to implement `defineConstants`.

Resolves #17.

### DIFF
--- a/packages/webpack/README.md
+++ b/packages/webpack/README.md
@@ -41,6 +41,12 @@ Takes an array of config setters (the functions returned by invoked webpack bloc
 
 Combines an array of blocks to a new joined block. Running this single block has the same effect as running all source blocks.
 
+#### defineConstants(constants: object): Function
+
+Replaces constants in your source code with a value (`process.env.NODE_ENV` for example) using the [webpack.DefinePlugin](https://webpack.github.io/docs/list-of-plugins.html#defineplugin). Pass an object containing your constant definitions: `{ [constantName: string]: <constantValue: any> }`.
+
+Special feature: Using `defineConstants` multiple times results in a single DefinePlugin instance configured to do all the replacements.
+
 #### env(envName: string, configSetters: Function[]): Function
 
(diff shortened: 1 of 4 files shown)


eXon:
That's awesome! It's exactly what I had in mind. 

One thing I'm still wondering though. Because it is called `defineConstants` and not `replaceCode`, maybe the value you want to define should be stringified automatically don't you think?

You would be able to do:
```js
defineConstants({
    DEV: process.env.NODE_ENV !== 'production'
})
```

The `JSON.stringify` should be implicit in my mind. I don't think this should be used as a search&replace within your code for anything. I think it would cover 99% of the use case and make the usage better.

andywer:
@eXon Excellent point! :)

I actually did forget to `JSON.stringify()` the values quite a few times in the past...

Will add that 👍👍

andywer:
Constants' values are now automatically `JSON.stringify()`-ed! (see the [commit](https://github.com/andywer/webpack-blocks/pull/43/commits/886b5d3488585d39ba91320fcd3d452e009665b9))

eXon:
This is perfect 👍

PR #54, opened 2016-12-23
Add js:next as a target

mjsisley:
Maybe js:next should be added as a target for tree shaking (keep es2015 imports and exports, but transpile the rest).

### DIFF
--- a/packages/webpack/index.js
+++ b/packages/webpack/index.js
@@ -69,7 +69,8 @@ function createBaseConfig (context) {
     },
 
     resolve: {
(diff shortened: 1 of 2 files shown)


mjsisley:
Sorry typo, should be `mainFields: ['browser', 'module', 'jsnext:main', 'main']`. Quicker if you just include in a later commit if you think it is worth including. Closing.

andywer:
I guess it just concerns the `webpack2` package, since webpack 1.x does not support tree shaking anyway, right?

PR #59, opened 2016-12-29
Use webpack-merge 2.0, fix react hot loader

### LINKED ISSUE #56: Support for react-hot loader
Right now there is no support for the react-hot loader : https://github.com/gaearon/react-hot-loader

Basically what you need is the devServer enabled (already got the block for that) and the 'react-hot' loader before the 'babel-loader'.

Right now, I'm doing (first loader is babel in my case):
```js
module.exports.module.loaders[0].loaders.unshift('react-hot');
```

Is there a better way to do that in a non-hackish way?

eXon:
Fix #56 

All I've done is add unit test for reactHot and update webpack-merge to 2.0.0.

It's a big jump in the version, but it ensures merging the loaders is done correctly for react-hot. However, it looks like the loaders order is now changed a little bit. Not a big deal, but worth testing enough before pushing this.

### DIFF
--- a/packages/webpack-common/src/extractText.js
+++ b/packages/webpack-common/src/extractText.js
@@ -18,7 +18,7 @@ export {
  */
 function getLoaderConfigByType (context, webpackConfig, fileType) {
   const loaderConfig = webpackConfig.module.loaders.find(
-    (loader) => regexEqual(loader.test, context.fileType(fileType))
+    (loader) => loader.test === context.fileType(fileType)
   )
 
   if (loaderConfig) {
@@ -26,12 +26,6 @@ function getLoaderConfigByType (context, webpackConfig, fileType) {
   } else {
     throw new Error(`${fileType} loader could not be found in webpack config.`)
   }
-}
-
-function regexEqual (x, y) {
-  return (x instanceof RegExp) && (y instanceof RegExp) &&
-   (x.source === y.source) && (x.global === y.global) &&
-   (x.ignoreCase === y.ignoreCase) && (x.multiline === y.multiline)
 }
 
 /**
--- a/packages/webpack-common/src/extractText.js
+++ b/packages/webpack-common/src/extractText.js
@@ -18,7 +18,7 @@ export {
  */
 function getLoaderConfigByType (context, webpackConfig, fileType) {
   const loaderConfig = webpackConfig.module.loaders.find(
-    (loader) => loader.test === context.fileType(fileType)
+    (loader) => regexEqual(loader.test, context.fileType(fileType))
   )
 
   if (loaderConfig) {
@@ -26,6 +26,12 @@ function getLoaderConfigByType (context, webpackConfig, fileType) {
   } else {
     throw new Error(`${fileType} loader could not be found in webpack config.`)
   }
+}
+
+function regexEqual (x, y) {
+  return (x instanceof RegExp) && (y instanceof RegExp) &&
+   (x.source === y.source) && (x.global === y.global) &&
+   (x.ignoreCase === y.ignoreCase) && (x.multiline === y.multiline)
 }
 
 /**
--- a/packages/webpack2/__tests__/integration.test.js
+++ b/packages/webpack2/__tests__/integration.test.js
@@ -23,37 +23,37 @@ test('complete webpack config creation', (t) => {
   ])
 
   t.is(webpackConfig.module.loaders.length, 8)
-  t.deepEqual(webpackConfig.module.loaders[0], {
+  t.deepEqual(webpackConfig.module.loaders[7], {
     test: /\.(js|jsx)$/,
     exclude: [ /\/node_modules\// ],
     loaders: [ 'babel-loader?{"cacheDirectory":true}' ]
   })
-  t.deepEqual(webpackConfig.module.loaders[1], {
+  t.deepEqual(webpackConfig.module.loaders[6], {
     test: /\.(sass|scss)$/,
     loaders: [ 'style-loader', 'css-loader', 'sass-loader' ]
   })
-  t.deepEqual(webpackConfig.module.loaders[2], {
+  t.deepEqual(webpackConfig.module.loaders[0], {
     test: /\.css$/,
     exclude: [ /\/node_modules\// ],
     loaders: [ 'style-loader', 'css-loader?importLoaders=1&localIdentName=[name]--[local]--[hash:base64:5]&modules' ]
   })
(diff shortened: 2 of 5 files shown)


packages/webpack-common/src/extractText.js
```diff
@@ -18,7 +18,7 @@ export {
  */
 function getLoaderConfigByType (context, webpackConfig, fileType) {
   const loaderConfig = webpackConfig.module.loaders.find(
-    (loader) => loader.test === context.fileType(fileType)
+    (loader) => regexEqual(loader.test, context.fileType(fileType))
```
andywer: I don't think we have to do a real regex comparison. `context.fileType(fileType)` should return the same regex reference throughout the whole configuration process.

The only scenario I can think of where it makes a difference: If there are two loader configs matching two file types which have semantically equal regexs to match. But I think we should not even match those as well 😉

packages/webpack2/__tests__/integration.test.js
```diff
@@ -23,37 +23,37 @@ test('complete webpack config creation', (t) => {
   ])
 
   t.is(webpackConfig.module.loaders.length, 8)
-  t.deepEqual(webpackConfig.module.loaders[0], {
```
andywer: Ahh, if the order has changed here, we should re-order the assertions as well. Having the indexes in the order 7, 6, 0, 1, 2 really is confusing.

packages/webpack-common/src/extractText.js
andywer: Maybe we should add a comment here making that a little clearer...

andywer:
You have some good points here 🙂

Just need to adjust it a little bit.

packages/webpack-common/src/extractText.js
eXon: It is unfortunally not the case anymore when I updated webpack-merge to 2.0.0. A test was failing because it couldn't find the correct loader for text/css.

andywer: Hmm. I propose we first investigate why this is happening. Looks like an easy fix to deep-compare the regexes, but that might be a minor bug in webpack-merge. If so we should not try to fix it afterwards in webpack-blocks, but fix it in webpack-merge.

andywer:
Guess we have to look into that slight comparison issue a little more (see [comment in code](https://github.com/andywer/webpack-blocks/pull/59#discussion_r94155145)).

I cannot say when I will be able to have a deeper look into it myself, though. I am at Munich airport between two connecting flights right now and visiting my girlfriend's family for the next few days... But maybe I manage to find some time :)

andywer:
The bright side is, though: The integration tests did their job perfectly fine 👍😄

packages/webpack-common/src/extractText.js
eXon: I think it is because of the deep clone webpack-merge is doing behind the scene. In a strict functional way, it is expected that the Regex instance might be a different one with the same value. I'm not sure it is bad to use a more reliable way to compare regexes. I don't think we can expect two instances with the same value on different loaders. I don't think regex are seen as primitive types in javascript.

andywer: Is it really doing a deep clone of the regex (new regex with same value)? I wouldn't expect that... (Or: If it really works this way, maybe this is this the bug 😉)

eXon:
The problem when comparing with the regex instance instead of the value is it is order specific. Meaning if another block add a loader and then your block adds a block with the same regex, you don't know it will not use your instance but the existing one.

I honestly believe it's better to go for comparing the regex value. That way, it is more functional :)

andywer:
@eXon Let's discuss this :)

> Meaning if another block add a loader and then your block adds a loader with the same regex, you don't know it will not use your instance but the existing one.

If two loaders are going to match the same file extensions then they use the same `context.fileType('foo/bar')` returning the same regex instance, since that is what the `context.fileType` is for :) That's how we ensure referential integrity.

The problem about comparing the regex' values is not only that it's slower (I really don't care for this particular use case), but that regex' are bitches that are really hard to compare. If we don't assume referential integrity anymore and always deep-compare the regex' values we encounter a new problem: Regex synonyms. Consider `/\.jsx?/` vs `/\.js|\.jsx/`. Different values, but matching the same extensions.

Referential integrity makes comparing trivial and helps me sleep well at night, without having to fear people open issues, because a comparison does not work for their crazy edge-case 🙈😅

eXon:
@andywer good news! looks like it's not a problem anymore with the last version of webpack-merge 👍

eXon:
Looks like it's still not working. My node_modules folder was not up to date. I'll continue searching why the regex instance is not the same.

andywer:
I was just debugging #76 and tried updating `webpack-merge` as well.

**I found that the last webpack-merge version leaving the regular expressions intact was `1.0.0`**
What's really strange is that its [changelog](https://github.com/survivejs/webpack-merge/blob/master/CHANGELOG.md) does not even mention version `1.0.1` or its changes that broke the regexs. It lists `1.0.0` and `1.0.2`, but no `1.0.1` 😳

In the light of #34 I have got to say that I am constantly loosing faith in `webpack-merge` here... 😕

andywer:
I stand corrected: `webpack-merge v1.0.2` broke it and I even found out why:

They introduced a breaking change and just published as a hot fix version. [This commit](https://github.com/survivejs/webpack-merge/commit/fd39e5a3e02ce777968b818645255af2c27f5973) introduces `lodash.cloneDeep` to deep-clone any object it sees.

eXon:
When they are comparing the loader test regex, `webpack-merge` is using [`String(regex)`](https://github.com/survivejs/webpack-merge/blob/master/src/join-arrays-smart.js#L8).

Maybe it's good enough for us too?

andywer:
@eXon Yeah, I give in. Let's do it like that as well, but let's add a short comment explaining that we do it because of `webpack-merge` deep-cloning stuff 😉

andywer:
Created #77, since this was your hot reloading PR. We should start to take more care not to throw different independent features into the same PR. Otherwise we will have "spaghetti PRs" soon :)

eXon:
Well the fix for the hot-reload bug was to upgrade webpack-merge but it will do the job. Awesome thanks 👍

andywer:
@exon I wasn't even aware of that anymore. So can we close this PR then? Might be good if you check out the latest master and confirm that it works.

eXon:
Your PR diff is almost identical so I'll close this. Thanks for the fix!

PR #65, opened 2017-01-01
Correct the mixup of webpack and webpack2

dmitriz:
Add few inline comments and instructions

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,12 +34,20 @@ npm install --save-dev @webpack-blocks/webpack @webpack-blocks/babel6 ...
 
 ## Usage
 
-Create a webpack 2 config with Babel support, dev server, HMR and PostCSS autoprefixer:
+Create a webpack config with Babel support, dev server, HMR and PostCSS autoprefixer:
 
 ```js
-const { createConfig, defineConstants, env, entryPoint, setOutput, sourceMaps } = require('@webpack-blocks/webpack2')
+// webpack.config.js
+
+const { 
+  createConfig, 
+  entryPoint, 
+  setOutput, 
+  defineConstants, 
+  env, 
+  sourceMaps } = require('@webpack-blocks/webpack')
 const babel = require('@webpack-blocks/babel6')
-const devServer = require('@webpack-blocks/dev-server2')
+const devServer = require('@webpack-blocks/dev-server')
 const postcss = require('@webpack-blocks/postcss')
 const autoprefixer = require('autoprefixer')
 
@@ -53,6 +61,8 @@ module.exports = createConfig([
   defineConstants({
     'process.env.NODE_ENV': process.env.NODE_ENV
   }),
+
+  // use only if `NODE_ENV === 'development'`:
   env('development', [
     devServer(),
     devServer.proxy({
@@ -62,6 +72,7 @@ module.exports = createConfig([
   ])
 ])
 ```
+Now simply run `webpack` or `webpack-dev-server`.
 
 Wanna use CSS modules? No problem!
 


andywer:
Thanks for the feedback! That's always valuable.

But you would like to see the initial example use webpack 1? I thought showing webpack 2 config from the start would be a good idea, since webpack 1 will probably soon reach its end of life.

dmitriz:
You are welcome!
Thanks for the awesome product, it is really needed!

I understood it was a typo to jump to wp 2 after the installation line refering to wp 1.

Also webpack 2 is already mentioned down below in "switching to wp 2".
However, it still in beta and actually caused me personally problems the few times I've tried it. 

Even with wp 1, I had hard time to remember the configuration and even wrote
https://github.com/dmitriz/min-webpack so I can get a working config and usage example simply by single install command `npm i min-webpack`.

And just today I had to ask questions http://stackoverflow.com/questions/41413665/how-to-run-wepback-with-projects-in-subdirectories (might webpack-blocks give a answer? 😄 )?

Concerning switching to wp 2, is it actually possible to keep both versions running in parallel?

dmitriz:
BTW, if any changes are desired, I'll be happy to update, of course.

andywer:
Thank you, @dmitriz :)

Guess I should create a Stack Overflow account... 🙈😅

Yeah, might make sense to stick to webpack 1 by default for now, but only if there are real issues with webpack 2 (would be nice to know what didn't work there for you). You cannot use webpack 1 and 2 in parallel, since you cannot install them both (unless using dirty hacks).

I will have a deeper look into it later. Feel free to tell us about the webpack 2 issues you encountered :)

andywer:
PS: Concerning your SO question: Might work if you rename `local.js` to `index.js` or use `entryPoint('./local.js')`. (The webpack config file is in the `subdir` as well, is it?)

dmitriz:
> Guess I should create a Stack Overflow account... 🙈😅

You can't go wrong with that 😄 

> Yeah, might make sense to stick to webpack 1 by default for now, but only if there are real issues with webpack 2 (would be nice to know what didn't work there for you). 

It depends on your target audience. If these are folks with less wp experience, I suppose they just want to get it up and running quickly and move on, then a stable version would probably suit them to begin with, even it is not the latest cutting edge. 


> You cannot use webpack 1 and 2 in parallel, since you cannot install them both (unless using dirty hacks).

That is unfortunate, having an easy switch would allow people to have some safety feeling trying wp2, so you know if things go wrong, you can quickly switch back, without spending your day on re-configuring 😈 

If there is even `nvm`, why can't be a `wvm`? 😄 


> I will have a deeper look into it later. Feel free to tell us about the webpack 2 issues you encountered :)

Thanks, here you go: https://github.com/andywer/webpack-blocks/issues/68

andywer:
> It depends on your target audience. If these are folks with less wp experience, I suppose they just want to get it up and running quickly and move on, then a stable version would probably suit them to begin with, even it is not the latest cutting edge.

Got to think that through... 🤔

> If there is even nvm, why can't be a wvm? 😄

Oh dear lord! 😅 I mean you can just `rm -rf node_modules/{webpack,webpack-dev-server}` and `npm install` in the other version again. That should take < 1 minute.

And changing your webpack-blocks config is also a matter of seconds :)

SuperManitu:
As Webpack 2.2 is now officially released and will be the npm latest version from next week on, I see no reason to stick to webpack 1

andywer:
@jvanbruegge Yeah, I think so, too.

@dmitriz What do you think? Can we just close the PR or should we highlight the usage of webpack 2 in this example? Or do you still see any inconsistencies in the README concerning webpack 1.x / 2.x?

jvanbruegge:
I would change the npm install line to webpack2

andywer:
True!

- `npm install` should install `@webpack-blocks/webpack2`
- "switching to webpack 2" section might be misleading (-> maybe rename it to "Switching to webpack 2 from webpack 1.x"?)
- "group() (presets)" does not use webpack2 yet

Anything else?

andywer:
Closing this in favor of #101. Please review it if you guys find some time.

And thanks :)

Issue #68, opened 2017-01-02
Webpack 2 + PostCSS errors (Invalid configuration object ...)


dmitriz:
I have tested `webpack.config.js` exactly as in 
https://github.com/andywer/webpack-blocks#usage
which lead to errors:

```bash
± |master ✓| → npm run build

> webpack2-blocks-test@0.1.0 build /Users/dmitrizaitsev/Dropbox/Sandbox/webpack2-blocks-test
> webpack

Invalid configuration object. Webpack has been initialised using a configuration object that does not match the API schema.
 - configuration has an unknown property 'postcss'. These properties are valid:
   object { amd?, bail?, cache?, context?, dependencies?, devServer?, devtool?, entry, externals?, loader?, module?, name?, node?, output?, performance?, plugins?, profile?, recordsInputPath?, recordsOutputPath?, recordsPath?, resolve?, resolveLoader?, stats?, target?, watch?, watchOptions? }
   For typos: please correct them.
   For loader options: webpack 2 no longer allows custom properties in configuration.
     Loaders should be updated to allow passing options via loader options in module.rules.
     Until loaders are updated one can use the LoaderOptionsPlugin to pass these options to the loader:
     plugins: [
       new webpack.LoaderOptionsPlugin({
         // test: /\.xxx$/, // may apply this only for some modules
         options: {
           postcss: ...
         }
       })
     ]

npm ERR! Darwin 12.6.0
npm ERR! argv "/usr/local/bin/node" "/Users/dmitrizaitsev/.npm-global/bin/npm" "run" "build"
npm ERR! node v6.9.2
npm ERR! npm  v4.0.5
npm ERR! code ELIFECYCLE
npm ERR! webpack2-blocks-test@0.1.0 build: `webpack`
npm ERR! Exit status 1
npm ERR! 
npm ERR! Failed at the webpack2-blocks-test@0.1.0 build script 'webpack'.
npm ERR! Make sure you have the latest version of node.js and npm installed.
npm ERR! If you do, this is most likely a problem with the webpack2-blocks-test package,
npm ERR! not with npm itself.
npm ERR! Tell the author that this fails on your system:
npm ERR!     webpack
npm ERR! You can get information on how to open an issue for this project with:
npm ERR!     npm bugs webpack2-blocks-test
npm ERR! Or if that isn't available, you can get their info via:
npm ERR!     npm owner ls webpack2-blocks-test
npm ERR! There is likely additional logging output above.

[...]

PR #69, opened 2017-01-03
Webpack 2 compatible PostCSS block + improvements

nightgrey:
Hi folks!

So, like mentioned in #68, I also encountered errors while using the latest webpack2 (rc.3 as of this moment) with the postcss block. After researching and testing a lot of stuff, I identified two issues. I'm not even sure if this is something @andywer wants to have, but as I did this for a project of my own, I thought why not submitting this; it at least could help somebody using webpack 2.

**Issues with current postcss block**
1. In webpack 2, adding a postcss property to the root of the webpack configuration object is invalid (see http://javascriptplayground.com/blog/2016/10/moving-to-webpack-2/#configuring-loaders)
2. After changing to the new webpack 2 style of configuring loaders, there are still errors. In webpack 2, passing in plugins to the postcss-loader within the webpack configuration is currently broken (see https://github.com/postcss/postcss-loader/issues/125#issuecomment-257127969)

So I made a "postcss2" block where I changed several things. I'm not sure if all of these changes are like the maintainers of webpack-blocks want them to be, I'll be happy to adapt this! There definitely are points which need discussion. 

**Changes made in the postcss 2 block**
- Used webpack 2 syntax "rules" instead of "loaders". "Loaders" is the webpack 1 style of defining loaders, which is backwards-compatible, but AFAIK will be deprecated. IMHO, this should be changed for all webpack 2 block. I'll be happy to create a pull request for the other blocks if this is wanted. See http://javascriptplayground.com/blog/2016/10/moving-to-webpack-2/#moduleloaders--modulerules
- Because of the current issue with webpack 2 and postcss-loader, loading plugins within the webpack configuration (like described above; see https://github.com/postcss/postcss-loader/issues/125#issuecomment-257127969) doesn't work. I changed the README to reflect this and added an error when someone tries to add plugins to the block. I'm not sure if this is what you guys want, though. Any ideas to do this better? See https://github.com/nightgrey/webpack-blocks/blob/88da42683768fecfdb8df5c2f233a795e56dd5b2/packages/postcss2/index.js#L32
- I added the possibility to pass other postcss options, like specifiying a parser, stringifier or syntax. This is needed to, for example, parse the SCSS syntax. I'd be happy to do this for the current postcss block, too!
- I improved/corrected the JSDoc comments a bit. For example, https://github.com/andywer/webpack-blocks/blob/1744d2878b6fbeb3bbd479597ab2e9ff5f85cf48/packages/postcss/index.js#L12 isn't actually valid; it would've to be `|` instead of `,`. There are some other improvements and more details, too. I'd be happy to do this for the current postcss block, too!

So, yeah, what do you say? Like already said, some things can easily be added to the current PostCSS block and with some others, I'm not sure if there are better solutions. I'm happy about any comment, thoughts or criticism on this. 

Edit: --amend'ed/updated my commit. Forgot to run standard and had some extra spaces, oops. Fixed.

### DIFF
--- a/packages/postcss2/CHANGELOG.md
+++ b/packages/postcss2/CHANGELOG.md
@@ -1,0 +1,5 @@
+# @webpack-blocks/postcss2 - Changelog
+
+## 0.1.0
+
+Initial release.
--- a/packages/postcss2/README.md
+++ b/packages/postcss2/README.md
@@ -1,0 +1,67 @@
+# Webpack blocks - PostCSS for Webpack 2
+
+[![JavaScript Style Guide](https://img.shields.io/badge/code%20style-standard-brightgreen.svg)](http://standardjs.com/)
+[![NPM Version](https://img.shields.io/npm/v/@webpack-blocks/postcss.svg)](https://www.npmjs.com/package/@webpack-blocks/postcss)
+
+This is the `postcss` block providing PostCSS configuration.
+
+
+## Usage
+
+***IMPORTANT:** Loading postcss plugins inside the webpack 2 (as of webpack@2.2.0-rc.3) configuration does not work right now, which is why we currently use postcss.config.js. Reference: https://github.com/postcss/postcss-loader/issues/125#issuecomment-257127969*
+
+### webpack.config.js
+```js
+const { createConfig } = require('@webpack-blocks/webpack')
+const postcss = require('@webpack-blocks/postcss')
+
+module.exports = createConfig([
+  postcss([
+    // You'd normally pass in your plugins here. Please check the note about
+    // configuring PostCSS plugins above!
+  ])
+])
+```
+
+### postcss.config.js
+
+```js
+module.exports = {
+  plugins: {
+    'autoprefixer': {
+      { browsers: ['last 2 versions'] }
+    }
+  }
+})
+```
+
+### Specificying postcss-loader options
+
+If you want to specifiy `syntax`, `stringifier` or `parser`, f.e. for using `postcss-scss`, you can pass an options object as second argument.
+
+```js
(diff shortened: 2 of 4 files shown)


andywer:
Hey @nightgrey! Thanks for your PR.

I am still in Oxford, but will be back in Germany tomorrow and will then have an in-detail look at it :)

The postcss block needs some adjustments, but I'd rather not have a webpack 1 and a webpack 2 version of every block. I am still interested in your proposed changes, though. Let's look into it tomorrow :)

nightgrey:
> The postcss block needs some adjustments, but I'd rather not have a webpack 1 and a webpack 2 version of every block.

I completely understand - and agree that it's obviously better to have only one block; on the other hand, it's a bit confusing right now; there at least should be a note somewhere. In any case, I just want to point out that it wasn't clear to me that the postcss block is not compatible with webpack 2. _(Don't get me wrong, I'm not blaming you or anyone else working on webpack blocks! In fact, I think webpack blocks is a great idea and I love the easy implementation. Overall: awesome stuff!)_ If I'm not mistaken the block still works with the beta versions of webpack 2; so this occured rather recently. Also, webpack 2 is not a stable release, so such things are expected.

Also, that just gave me an idea.. I'm not sure if it works, though. Let me check and come back to you.

andywer:
You are right, though, as well. We need to have a better solution for this issue.

I thought about throwing an error if you try to set the PostCSS plugins from within the webpack config when using webpack v2 (see [#68](https://github.com/andywer/webpack-blocks/issues/68#issuecomment-269946672)).

We should be able to determine the webpack version by having a look at the `webpack` instance which is in the context object.

andywer:
Issue fixed by #74.

PR #71, opened 2017-01-04
Webpack2: Remove json-loader

### LINKED ISSUE #63: For webpack2, remove json-loader
Per the docs, it's no longer necessary for webpack2: https://webpack.js.org/guides/migrating/#json-loader-is-not-required-anymore

andywer:
Webpack 2 comes with a default `json-loader` config. We will just use that by default.

Resolves #63. Read that issue for details.


andywer:
Reminder: Must publish the package after merging.

andywer:
Published as `@webpack-blocks/webpack2 v0.3.1`.

PR #78, opened 2017-01-11
Composable dev server

andywer:
Using multiple `devServer(<options>)` in one config now works as expected.

Also improved its tests.


andywer:
Note: Must eventually release a new version of `dev-server`, `dev-server2` and `webpack-common`.

PR #79, opened 2017-01-11
Slightly changed `invokeConfigSetters` code

andywer:
I hope that makes the `invokeConfigSetters` code and its usage a little easier to understand.

Also fixes a minor issue I encountered when using `group()`/`env()`:
The webpack config object that was passed to blocks as 2nd parameter contained duplicate prop values when blocks where used in a `group()`/`env()`. Did not cause any bugs yet, though AFAIK.

@eXon Did I succeed to improve the code's readability or didn't I? 😅


eXon:
Yes a lot easier to understand what's going on 👍

PR #86, opened 2017-01-17
add webpack typescript block

SuperManitu:
Adds a typescript block for compiling typescript and typescript react (.ts and .tsx) with webpack.
Tests are passing.

### DIFF
--- a/packages/typescript/README.md
+++ b/packages/typescript/README.md
@@ -1,0 +1,29 @@
+# Webpack blocks - SASS
+
+[![JavaScript Style Guide](https://img.shields.io/badge/code%20style-standard-brightgreen.svg)](http://standardjs.com/)
+
+This is the `typescript` block providing Typescript support for webpack. Uses `awesome-typescript-loader`.
+
+
+## Usage
+
+```js
+const { createConfig } = require('@webpack-blocks/webpack')
+const typescript = require('@webpack-blocks/typescript')
+
+module.exports = createConfig([
+  typescript()
+])
+```
+
+## Options
+
+Uses the default tsconfig.json in the root directory to pass options (See [here](https://github.com/s-panferov/awesome-typescript-loader#tsconfigjson))
+
+## Webpack blocks
+
+Check out the
+
+👉 [Main Documentation](https://github.com/andywer/webpack-blocks)
+
+Released under the terms of the MIT license.
--- a/packages/typescript/README.md
+++ b/packages/typescript/README.md
@@ -1,8 +1,8 @@
-# Webpack blocks - SASS
+# Webpack blocks - TypeScript
 
 [![JavaScript Style Guide](https://img.shields.io/badge/code%20style-standard-brightgreen.svg)](http://standardjs.com/)
 
-This is the `typescript` block providing Typescript support for webpack. Uses `awesome-typescript-loader`.
+This is the `typescript` block providing TypeScript support for webpack. Uses `awesome-typescript-loader`.
 
 
 ## Usage
--- a/packages/typescript/index.js
+++ b/packages/typescript/index.js
@@ -9,13 +9,13 @@ const { CheckerPlugin, TsConfigPathsPlugin } = require('awesome-typescript-loade
 module.exports = typescript
 
 /**
- * @param {object}   [advancedPathResolution]                  See https://github.com/s-panferov/awesome-typescript-loader#advanced-path-resolution-in-typescript-20
+ * @param {object} [advancedPathResolution]                  See https://github.com/s-panferov/awesome-typescript-loader#advanced-path-resolution-in-typescript-20
  * @param {object} [options.tsconfig]
  * @param {object} [options.compiler]
  * @return {Function}
  */
 function typescript (options) {
-  return (context) => ({
+  const setter = (context) => ({
     resolve: {
       extensions: ['.ts', '.tsx']
     },
@@ -29,6 +29,22 @@ function typescript (options) {
         }
       ]
     },
-    plugins: options ? [new CheckerPlugin(), new TsConfigPathsPlugin({ tsconfig: options.tsconfig, compiler: options.compiler })] : [new CheckerPlugin()]
+    plugins: [
+      new CheckerPlugin()
+    ].concat(
+      options ? [
+        new TsConfigPathsPlugin({ tsconfig: options.tsconfig, compiler: options.compiler })
+      ] : []
+    )
   })
+  
+  return Object.assign(setter, { pre })
+}
+
+function pre (context) {
+  if ('application/x-typescript' in context.fileType.all()) {
+    // file type is already there
+  } else {
+    context.fileType.add('application/x-typescript', /\.(ts|tsx)$/)
+  }
 }
--- a/packages/typescript/index.js
+++ b/packages/typescript/index.js
@@ -1,0 +1,34 @@
+/**
+ * Typescript webpack block.
+ *
+ * @see https://github.com/s-panferov/awesome-typescript-loader
+ */
+
+const { CheckerPlugin, TsConfigPathsPlugin } = require('awesome-typescript-loader')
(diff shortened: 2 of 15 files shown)


andywer:
Awesome! Thanks, @SuperManitu! 🎉🎉

I might add an end-to-end test for the typescript block, but this is great. In the very beginning of the project I already thought about a typescript block, but since I haven't been actively using TS together with webpack I wasn't quite sure which loader to take 😉

andywer:
Also opened #87, since I feel that MIME type adding might become a rather frequent issue.

SuperManitu:
I tried modifying the test app, but without the block being on npm this is rather hard.
Im using the awesome-typescript-loader for a while now and it's the best one out there.

Also, you are welcome 😄

andywer:
Small tip when testing unpublished packages:

First, `cd unpublished-package/` and `npm link`. Then `cd test-app/` and `npm link name-of-unpublished-package`. Ready to go 😉

andywer:
Disclaimer: Might or might not work on Windows (*"Schrödinger's operating system"*)

SuperManitu:
Will try to modify the test app
Work on Linux anyway :D

SuperManitu:
OK, I got it working, but you need (obviously) the npm link before. Should I commit the modified sample app or should I create a new branch with PR to be merged once this is is merged?

andywer:
Updating the test app in a separate PR would be desirable, since it is dependant on another, but not tightly coupled.

Will merge this one tomorrow!

andywer:
Just noticed: Please add the `typescript` block to the list of blocks in the main README :)

Would have done it myself, but it seems I can only edit files in your feature branch that you already modified.

andywer:
Took the liberty to add a pre hook to the block. Feel free to review my commit :)

## Left to do

- [ ] Write end-to-end test (for webpack and webpack2; should be copy & paste)

I can write the test myself if you want, probably later today. But feel free to write it yourself if you feel like it.

packages/typescript/index.js
```diff
@@ -29,6 +29,22 @@ function typescript (options) {
...
+        new TsConfigPathsPlugin({ tsconfig: options.tsconfig, compiler: options.compiler })
+      ] : []
+    )
   })
+  
+  return Object.assign(setter, { pre })
+}
+
+function pre (context) {
+  if ('application/x-typescript' in context.fileType.all()) {
```
SuperManitu: I would not leave an empty if branch, I would use:
`if (!('application/x-typescript' in context.fileType.all())) {`
and just have the else branch

andywer: True, but the `if (!(quite~long~thing~here()))` looks shitty as well... How about:

```js
const registeredTypes = context.fileType.all()
if (!('application/x-typescript' in registeredTypes)) {
  ...
}
```

SuperManitu: yes, reads very natural: if not typescript in registeredTypes 😃

SuperManitu:
Is this the end to end test you meant?

andywer:
No, this is what I call a unit test in this context. The e2e tests are in `webpack` and `webpack2`, they actually run webpack with a given configuration and check that the build artifacts are fine.

But having a unit test is not wrong, though.

I should probably add documentation for the test concept somewhere...

SuperManitu:
ok, I get an error with webpack 1: `error: window not defined`, webpack2 works as expected (when linking the block)

andywer:
Ehhmm, I think it had something to do with the `style-loader`, `css-loader` and/or `ExtractTextPlugin`... In your e2e test?

packages/webpack/__e2e-fixtures__/typescript/webpack.config.js
```diff
@@ -0,0 +1,32 @@
...
+    path.join(__dirname, 'build/bundle.js')
+  ),
+  typescript(),
+  performance({
+    maxAssetSize: 100000,
+    maxEntrypointSize: 500000,
+    hints: 'error'
+  }),
+  addPlugins([
+    new webpack.DefinePlugin({
```
andywer: I'd say we replace that by `defineConstants()`. But since this is already tested by the babel/postcss/extract-text e2e test I would just throw out the `process.env` injection thing altogether 😉

SuperManitu: i think it is a copy & paste error, copied the babel config again, now i get `you need an approriate loader`

SuperManitu:
Is there a difference in the loader config between webpack 1 and 2?

andywer:
No, there shouldn't be any. We are using the webpack 1.x loader syntax for both, webpack 1 and 2, since webpack 2 is backward-compatible.

Suggestion: `console.log(module.exports.loaders)` in your webpack config and have a look at what is eventually generated. Maybe the loader's `test` regex is flawed.

SuperManitu:
ok, I imported the wrong module as typescript
still get the windows is not defined error

packages/webpack/__e2e-fixtures__/typescript/styles.css
```diff
@@ -0,0 +1,5 @@
+$MARGIN: 40px;
```
andywer: Do these styles make sense at all? There is no SASS or similar loader defined to resolve the variable.

Suggestion: Drop the stylesheet completely 😉

packages/webpack2/__e2e-fixtures__/typescript/webpack.config.js
```diff
@@ -0,0 +1,32 @@
...
+    path.join(__dirname, 'build/bundle.js')
+  ),
+  typescript(),
+  performance({
+    maxAssetSize: 100000,
+    maxEntrypointSize: 500000,
+    hints: 'error'
+  }),
+  addPlugins([
+    new webpack.DefinePlugin({
```
andywer: This test should use `definePlugins()` as well. Just recognized that the babel/postcss/extract-text test used `defineConstants()` in `webpack`, but `DefinePlugin` in `webpack2`. Will change it there.

andywer:
As I commented in the file it might be a good idea to drop the stylesheet in the e2e test altogether. It isn't really used and already covered by another e2e test.

Maybe that will make the `window is not defined` error vanish already.

SuperManitu:
this should fix the texts, I removed the css

andywer:
I think you still need to explicitly copy the `dependencies` of `@webpack-blocks/typescript` into the `devDependencies` of `@webpack-blocks/webpack`/`@webpack-blocks/webpack2`. Tests should work then :)

andywer:
(FYI: `lerna bootstrap` is automatically run which will symlink the webpack-block packages' cross-dependencies. So it does not care if it has been published or not, since it will symlink them anyway 😉)

SuperManitu:
Ok, didn't know that, fixed now

andywer:
No, you couldn't know. But I added an issue to document that: #91 😉

SuperManitu:
Anything left to do?

andywer:
Sorry, will have to wait until tomorrow.

packages/typescript/README.md
```diff
@@ -1,8 +1,8 @@
-# Webpack blocks - SASS
+# Webpack blocks - Typescript
 
 [![JavaScript Style Guide](https://img.shields.io/badge/code%20style-standard-brightgreen.svg)](http://standardjs.com/)
 
-This is the `typescript` block providing Typescript support for webpack. Uses `awesome-typescript-loader`.
+This is the `typescript` block providing typescript support for webpack. Uses `awesome-typescript-loader`.
```
andywer: Just one suggestion: I think it's written "TypeScript" (https://en.wikipedia.org/wiki/TypeScript). But it's really no big deal 😉

andywer:
Published! :)

And thanks again for the great work 🎉

SuperManitu:
You are welcome :)

PR #88, opened 2017-01-18
Modify test-app to check typescript build too

jvanbruegge:
ONLY MERGE AFTER #86 HAS BEEN PUBLISHED ON NPM

Adds a TsApp React component written in TSX, adds the typescript block to the webpack config.
Tests will fail initially, should pass once the typescript block is on npm.


test-app/package.json
```diff
@@ -7,8 +7,11 @@
     "node": ">= 5.0"
   },
   "devDependencies": {
+    "@types/node": "^7.0.0",
+    "@types/react": "^15.0.0",
     "@webpack-blocks/babel6": "^0.x",
     "@webpack-blocks/css-modules": "^0.x",
+    "@webpack-blocks/typescript": "^0.x",
```
andywer: Please change to a proper version number. I already changed it for `babel6` and `css-modules` in the release branch :)

test-app/src/TsApp.tsx
```diff
@@ -0,0 +1,9 @@
+import * as React from 'react'
+const styles = require('./App.css')
+
+const TsApp : (() => JSX.Element) = () =>
```
andywer: I suggest we go all the way here. If we have a TypeScript-based example then let's drop the JS one.
And we can rename `TsApp` => `App` then 😉

andywer:
A `git fetch && git rebase origin/typescript` might be a good idea here. It's based on the other PR, but not yet up-to-date in respect of its recent changes 😉

SuperManitu:
rebased on the current master

andywer:
One question: I tried it locally and it works nicely. But it creates a `App.js` and `App.js.map` file in `test-app/src/`. Can we make it only write those to the output bundle instead of the `src/` directory? Or at least in a `lib/` directory?

jvanbruegge:
This should only happen if you run tsc over the directory. For me running `npm run build` just updates the bundle.js (and the css/html)

andywer:
Ahh, sorry, nevermind! Just installed `atom-typescript` and it has this "compileOnSave" feature enabled by default 🙈😉

jvanbruegge:
you can add `"compileOnSave": false` to the `tsconfig.json`

andywer:
Yeah, that looks like a good idea! 👍

andywer:
Did you write the `tsconfig.json` by hand or is it generated by some other tool? (Since it's the only file with 4 spaces indent)

jvanbruegge:
Basicly copy & paste it over and over, my default indent is 4 spaces. It's no problem to change that

andywer:
Wow, that was fast 😉

jvanbruegge:
Just edited in Github

andywer:
I thought so, but still... Will merge it after Travis is done running the tests :)

jvanbruegge:
Cool 😄

PR #92, opened 2017-01-19
add tslint block

jvanbruegge:
for all typescript linting needs :)

### DIFF
--- a/packages/tslint/index.js
+++ b/packages/tslint/index.js
@@ -18,7 +18,7 @@ function tslint (options) {
     loaders: [ 'tslint-loader' ]
   })
 
-  const module = context => (context.webpack.version === 1 ? {
+  const module = context => (context.webpackVersion.major === 1 ? {
     preLoaders: [ loader(context) ]
   } : {
     loaders: [ Object.assign({}, loader(context), {
@@ -28,14 +28,14 @@ function tslint (options) {
 
   const setter = (context) => Object.assign({
     module: module(context),
-    plugins: context.webpack.version === 2 ? [
+    plugins: context.webpackVersion.major === 2 ? [
       new context.webpack.LoaderOptionsPlugin({
         options: {
           tslint: options
         }
       })
     ] : []
-  }, context.webpack.version === 1 ? { tslint: options } : undefined)
+  }, context.webpackVersion.major === 1 ? { tslint: options } : undefined)
 
   return Object.assign(setter, { pre })
 }
--- a/packages/tslint/index.js
+++ b/packages/tslint/index.js
@@ -18,7 +18,7 @@ function tslint (options) {
     loaders: [ 'tslint-loader' ]
   })
 
-  const module = context => (context.webpack.getVersion() === 1 ? {
+  const module = context => (context.webpack.version === 1 ? {
     preLoaders: [ loader(context) ]
   } : {
     loaders: [ Object.assign({}, loader(context), {
@@ -28,14 +28,14 @@ function tslint (options) {
 
   const setter = (context) => Object.assign({
     module: module(context),
-    plugins: context.webpack.getVersion() === 2 ? [
+    plugins: context.webpack.version === 2 ? [
       new context.webpack.LoaderOptionsPlugin({
         options: {
           tslint: options
         }
       })
     ] : []
-  }, context.webpack.getVersion() === 1 ? { tslint: options } : undefined)
+  }, context.webpack.version === 1 ? { tslint: options } : undefined)
 
   return Object.assign(setter, { pre })
 }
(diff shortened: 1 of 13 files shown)


andywer:
Awesome! :)

I am on a train right now and writing from my iPhone, but if I should forget to write a main contributors list please tell me. You are definitely on it 😉

andywer:
Feel free to add the block to the list in the main README as well :)

andywer:
I applied two minor changes. Added an npm version badge (which will work as soon as it's published) and I replaced the `require('webpack')` by using `context.webpack`.

SuperManitu:
Is the options plugin also working for webpack 1? Maybe I should add an e2e test

andywer:
I was going to write this just now 😉 Could you please use the `tslint` block in the TypeScript e2e tests?

And of course, the `LoaderOptionsPlugin` will probably work with webpack 2 only... We might just copy the approach used in the [postcss block](https://github.com/andywer/webpack-blocks/blob/master/packages/postcss/index.js#L46) for now.

But that reminds me of something... I wanted to add the webpack version to the `context`. Need to open an issue for that as well.

SuperManitu:
if we have the version we dont need to split, becaude then the object is dumped in the webpack config

andywer:
But for webpack 1.x we will need to add the tslint config directly to the webpack config object, for webpack 2 we need to use the `LoaderOptionsPlugin`. So we need the version to distinguish which way to go, but it must be able to do both, right?

SuperManitu:
the block yes

SuperManitu:
I can do this when #96 is merged

andywer:
Sure, that would be the easiest solution :)

We just have to add a peer dependency `'@webpack-blocks/core': '>= 0.4.0'` to the tslint block then.

SuperManitu:
I rebased on #96, but the tslint-loader does not complain even if it should (mixed semicolon style in app.ts). Works in both webpack version now

SuperManitu:
The block is now completely working, you can test it by adding a semicolon in the app.ts of the end2end test. It will break the build

jvanbruegge:
@andywer this should be finished now, can you take another look?

andywer:
I think this feature branch is quite broken... There are still some remains of code that is not part of the `tslint` block, but not part of the `master` branch either.

Have a look at the changes of `webpack/index.js` and `webpack2/index.js`.
You merged some other feature branch into this one before, I think that in combination with the latest rebasing broke it. Sorry 🙈

jvanbruegge:
You are right, one wrong commit sneaked in. I deleted it.

packages/tslint/index.js
```diff
@@ -0,0 +1,48 @@
...
+ */
+function tslint (options) {
+  options = options || {}
+
+  const loader = context => ({
+    test: context.fileType('application/x-typescript'),
+    loaders: [ 'tslint-loader' ]
+  })
+
+  const module = context => (context.webpackVersion.major === 1 ? {
```
andywer: `standard` usually requires you to put a single parameter of an arrow function into parenthesis. Maybe the linting of this code does not work? 🤔

packages/tslint/index.js
```diff
@@ -0,0 +1,48 @@
...
+    preLoaders: [ loader(context) ]
+  } : {
+    loaders: [ Object.assign({}, loader(context), {
```
andywer: Might be easier readable as `loaders: [ loader(context, { enforce: 'pre' }) ]`. What do you think?

andywer: (2nd param for `loader()`, will be merged into the loader object using `Object.assign()`)

packages/tslint/index.js
jvanbruegge: I think I remember putting a semicolon there, and it was cought. Can check later

jvanbruegge:
I tested it, standard is indeed run on the index.js file. Adding a semicolon throws an error.
I also updated the loader function, indeed it makes it a lot more readable

andywer:
Cool 👍

And I just recognized they relaxed the arrow-function-must-have-parenthesis rule in `standard` 😉 (feross/standard#414)

andywer:
Just changed a line in favor of more consistent code style. We should stick to one style (single-param arrow functions have parenthesis or they don't) at least within the same file 😉

jvanbruegge:
Ok

andywer:
Off-topic: Since you live in Munich... Do you know https://twitter.com/munichjs? :)

jvanbruegge:
No, but it looks very interesting, especially this meetup at the google office

andywer:
Yeah, the people attending it seemed quite interesting as well. Just stumbled across and thought you might have already been there.

andywer:
Published!

jvanbruegge:
yay 🎉 🎉 🎉

Issue #93, opened 2017-01-20
Add webpack version to context


andywer:
The `context` should contain the webpack version so blocks that need to configure loaders/plugins differently for webpack 1 and webpack 2 can rely on that version.

The `postcss` block does this kind of differentiation already, but in a quite hacky way. And it seems that the `tslint` block will need to do something similar (see #92).

PR #96, opened 2017-01-20
add version property on webpack object

SuperManitu:
This is a very simple and convenient solution for #93

### DIFF
--- a/packages/core/src/index.js
+++ b/packages/core/src/index.js
@@ -18,7 +18,7 @@ const isFunction = (value) => typeof value === 'function'
  * @param {Function[]} configSetters  Array of functions as returned by webpack blocks.
  * @return {object}                   Webpack config object.
  */
-function createConfig (webpack, configSetters) {
+function createConfig (webpack, versionString, configSetters) {
   if (!webpack) {
     throw new Error(`No webpack instance passed.`)
   }
@@ -26,14 +26,37 @@ function createConfig (webpack, configSetters) {
     throw new Error(`Expected parameter 'configSetters' to be an array of functions.`)
   }
 
+  const webpackVersion = parseVersionString(versionString)
+
   const fileType = createFileTypesMapping(defaultFileTypes)
-  const context = { fileType, webpack }
+  const context = { fileType, webpackVersion, webpack }
 
   invokePreHooks(configSetters, context)
   const config = invokeConfigSetters(configSetters, context)
   const postProcessedConfig = invokePostHooks(configSetters, context, config)
 
   return postProcessedConfig
+}
+
+/**
+ * Parses a semver string into an object
+ * @param {string} string
+ * @return {object}
+ */
+function parseVersionString (string) {
+  const semver = string.split('-')
+  const versions = semver[0]
+    .replace(/[^0-9.]/g, '')
+    .split('.')
+    .map(parseInt)
+
+  return {
+    major: versions[0],
+    minor: versions[1],
+    patch: versions[2],
+    suffix: semver[1],
+    string: string
+  }
 }
 
 /**
--- a/packages/core/src/__tests__/createConfig.test.js
+++ b/packages/core/src/__tests__/createConfig.test.js
@@ -3,6 +3,7 @@ import sinon from 'sinon'
 import { createConfig } from '../index'
 
 const webpack = {}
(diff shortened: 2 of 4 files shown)


andywer:
Thanks for the PR!

The solution you propose is straight-forward and I agree that `webpack.version` would be ideal (maybe the webpack authors should have added this already), but I don't think we should monkeypatch the webpack instance at all.

Imagine they add a `version` property tomorrow which has a different format...

`context.webpackVersion` might not look as nice, but such an approach may safe us a lot of future trouble 😉

I would also propose we store the parsed semver read from webpack's `package.json` instead of a hard-coded major version integer.

What do you think? Will update the PR (if I can) or create another one. Unless you disagree with my proposals. Let's settle on the "how" first :)

SuperManitu:
sure, maybe even as object, so i can compare webpack.webpackVersion.major. Something like this:
```
webpackVersion: {
  major: 2
  minor: 1
  patch: 0
  suffix: rc.0
  string: "2.1.0-rc.0"
}
```

andywer:
Exactly! I would say we use https://github.com/npm/node-semver, since it's a quite small dependency and the same parser `npm` uses :)

SuperManitu:
I dont think we need an extra dependency, node-semver does a lot more than we need, I updated the core, but the tests in webpack and webpack2 are breaking without me finding the reason

andywer:
True, `node-semver` does stuff we don't need as well. I just fear that "re-inventing the wheel" might be error-prone here, since semver allows many kinds of rare edge cases, I think.

For example the `.map(parseInt)` line: Makes sense for all versions I can remember, but I am not sure if I would go as far as saying "there will always be an integer". You know...

One more thing: `createConfig()` having two parameters before the actual setters doesn't seem nice (esp. if we add even more in future). I'd like to propose turning the first parameter into an object (`initialContext`), containing `webpack` and `webpackVersion`.

SuperManitu:
The problem with that is that this would break all blocks that use context.webpack

andywer:
It would not break any blocks. I am just talking about how the stuff is passed to `createConfig()`, the `context` will still look the same 😉

Like that:
```js
function createConfig (initialContext, configSetters) {
  // ...
  const context = Object.assign({}, initialContext, { fileType })
  // ...
}
```

SuperManitu:
Ahh, ok

andywer:
Ahh, screw my `node-semver` argumentation... After thinking about it some more I am not even sure about anymore 🙈😅

Custom version parsing code might be fine if written in a reliable way. `node-semver` is maybe not that well suited for our use case.

But I have some rather small comments on the current implementation. Will comment the code.

packages/core/src/index.js
```diff
@@ -37,6 +39,27 @@ function createConfig (webpack, configSetters) {
...
 
 /**
+ * Parses a semver string into an object
+ * @param {string} string
+ * @return {object}
+ */
+function parseVersionString (string) {
+  const semver = string.split('-')
+  const versions = semver[0]
+    .replace(/[^0-9.]/g, '')
```
andywer: Why would you `.replace(/[^0-9.]/g, '')`? That means cutting off the major version, doesn't it? 🤔

packages/core/src/index.js
```diff
@@ -37,6 +39,27 @@ function createConfig (webpack, configSetters) {
 }
 
 /**
+ * Parses a semver string into an object
+ * @param {string} string
+ * @return {object}
+ */
+function parseVersionString (string) {
+  const semver = string.split('-')
```
andywer: `const [ release, prerelease ] = string.split('-')` might be better readable

packages/core/src/index.js
```diff
@@ -37,6 +39,27 @@ function createConfig (webpack, configSetters) {
 }
 
 /**
```
andywer: I think we should move the version-parsing function into a separate file. It's not big, but it's also not related to the rest of the file's code, it's something completely independent.

andywer:
Would you mind if I start a new feature branch based on what we discussed today?

Alternatively we can also change the code here and we squash everything before merging, so there are not so many back-and-forth commits.

SuperManitu:
Yeah, it's easier to start a new one

PR #99, opened 2017-01-22
Add webpackVersion to context

### LINKED ISSUE #93: Add webpack version to context

andywer:
This PR is the successor of #96 (thanks, @jvanbruegge!). As discussed this PR comes with a few changes (`parseVersion()` slightly changed and in its own file, more tests)

Resolves #93.

### DIFF
--- a/packages/core/CHANGELOG.md
+++ b/packages/core/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Next release
 
+- Changed signature of `createConfig()` to allow passing `context.webpackVersion`
 - Added `text/x-less` file type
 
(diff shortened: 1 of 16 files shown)


andywer:
@jvanbruegge I hope you did not start working on it as well yet. Review it if you find some time :)

jvanbruegge:
No I did not :)

PR #106, opened 2017-01-31
Fix typo

azat-io:


andywer:
Thanks for the PR, @azat-io! :)

I just restarted the failed build. Might be one of those random incomplete `npm install` people complain about...

andywer:
The problem seems to be a broken `ExtractTextPlugin` version: webpack-contrib/extract-text-webpack-plugin#381. It has already been fixed, but the fix has not been published yet.

PR #112, opened 2017-02-10
Fix optionalness of plugins param

sapegin:
**Current behavior**

This:

```js
createConfig([
  postcss(),
])
```

produces a warning “PostCSS Config could not be loaded. Please check your PostCSS Config”.

But this:

```js
createConfig([
  postcss([]),  // Empty plugins array
])
```

works fine.

This pull request removes the warning.

### DIFF
--- a/packages/postcss/index.js
+++ b/packages/postcss/index.js
@@ -16,6 +16,7 @@ module.exports = postcss
  * @return {Function}
  */
 function postcss (plugins, options) {
+  plugins = plugins || []
   options = options || {}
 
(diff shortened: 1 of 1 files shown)


andywer:
Ahhh, I see. For a moment I still thought "Why would you use the PostCSS loader, but with not plugins?". But there is the PostCSS config file now...

Excellent 👍

Thx, @sapegin! :)

sapegin:
And you don’t have CSS block anyway ;-)

andywer:
Published as `@webpack-blocks/postcss v0.4.1`

PR #118, opened 2017-02-23
Add package keywords for npm discoverability

adamdicarlo:
Hi @andywer! Webpack Blocks is such an awesome idea. I've found it a bit difficult to find the various available blocks on npmjs.com though, so here are some keywords! :)

### DIFF
--- a/packages/babel6/package.json
+++ b/packages/babel6/package.json
@@ -12,6 +12,10 @@
   "engines": {
     "node": ">= 4.0"
   },
+  "keywords": [
+    "webpack",
+    "webpack-blocks"
+  ],
   "repository": "andywer/webpack-blocks",
   "bugs": "https://github.com/andywer/webpack-blocks/issues",
   "dependencies": {
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -12,6 +12,7 @@
   },
   "keywords": [
     "webpack",
+    "webpack-blocks",
     "block",
     "feature",
     "functional"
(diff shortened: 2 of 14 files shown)


andywer:
Thanks, @adamdicarlo! 👍

There is a list of the available blocks in the [README](https://github.com/andywer/webpack-blocks#available-webpack-blocks), though. Is it because you rather like to use `npm search` or is the list in the README not placed prominently enough?

andywer:
(Changes will be visible on npm after the next release)

adamdicarlo:
Hmm, true! I did see that, although wasn't focusing enough to realize it's supposed to be an exhaustive list, I think. Though I probably would have also wondered what else might be out there that wasn't listed in this README. And there are a few at least -- I sent a PR over to `webpack-blocks-purescript` to add a keyword there too, and that's published (the author also added the keywords to his couple other packages and published 'em!). Great to have multiple avenues of discovery!

I came to know about webpack-blocks via the [awesome webpack](https://github.com/webpack-contrib/awesome-webpack) list btw!

adamdicarlo:
Oh right, I remember—part of the reason I was on npmjs.com after looking at the readme: I wanted to see the package pages to see how popular the packages were.

andywer:
Interesting! :)

Yeah, I saw the ecliptic webpack-blocks packages as well. I might contact them. Already thought about a minimalistic website listing all available blocks, but I have got more important things to handle right now, so maybe later.

Just hard to tell the popularity, since npm does still not support download stats for scoped packages...

adamdicarlo:
Oh, I didn't know that about scoped packages! Hope they get that working soon.

andywer:
Here is the issue, in case you want to track it yourself 😉

npm/registry#59

PR #119, opened 2017-02-25
Fix strange PostCSS/SASS issue

### LINKED ISSUE #116: PostCSS block with SASS and sourceMap
Adding `sass({sourceMap: true})` produces error:

```
TypeError: Path must be a string. Received undefined
```

andywer:
Fixes #116. Introduces a regression test as well.


andywer:
Published as `@webpack-blocks/postcss v0.4.2`

PR #123, opened 2017-02-27
Remove webpack 2.0 elements for 1.0 branch

aaronjensen:
This branch is only for webpack 1.0 now.


aaronjensen:
I was able to remove webpack-common entirely. We'll remove it from the 2.0 branch as well.

In the process, I converted some es6 imports to commonjs requires. It didn't seem worth it to have babel in the pipeline just for es6 imports. It actually bit me a few times trying to run tests (tests didn't build first). Any way, I hope that's good w/ you.

andywer:
Yeah, I forgot to tell you that. Thanks for removing the `webpack-common`! 👍👍

andywer:
Build is failing for node 4 now, though. It's features like default parameters that don't work in node 4 without Babel.

aaronjensen:
> Build is failing for node 4 now, though. It's features like default parameters that don't work in node 4 without Babel.

Ah, indeed. Are you still interested in supporting Node 4? I can fix up those issues if you'd like or bump the minimum versions to Node 6. Given that this is typically run locally (it's not a server library) I'd feel less bad about only supporting Node 6. Node 4 is only about a month away from maintenance mode. Up to you though.

andywer:
Maintenance mode in one month is a good catch, but nevertheless I would stick to node 4 support at least until node 8 is out (supporting last two LTS releases seems reasonable).

aaronjensen:
Updated to support node 4. I checked and webpack still supports node 4, so it's not a bad idea even if not having array destructuring and defaults is painful :)

That said, I'm aware that this branch is on hold until we determine the support strategy for v.next

Issue #14, opened 2016-12-07
Support for happypack


amilajack:
[happypack](https://github.com/amireh/happypack) allows webpack to be parallelized across threads. Adding support for it and webpack dll's would be a huge leap in terms of build speeds.

Willing to make a PR for this with the help of @amireh

Thoughts?

andywer:
Hi @amilajack (and @amireh). Thanks for your interest to contribute!

How about writing the block in the happypack package first (like in `happypack/webpack-block.js`): You can already publish and use it and then open a PR here when you think it works / it's stable?

I definitely see the benefit, we just have to come up with some kind of process here :)

amireh:
Of course. I'd be happy to help. I should point out though that this is the first time I see webpack-blocks (nice concept :smile:) so I don't have enough context.

I feel that adding the block definition in the main repository seems like the wrong spot, shouldn't webpack-blocks be the one concerned in providing support for it?

Also keep in mind that (unfortunately) HappyPack is a strange beast, somewhat like the extract-text-plugin. For an API, I would imagine _decorating_ a block with happypack support, like:

```javascript
const happyPackBlock = require('@webpack-blocks/happypack');

module.exports = createConfig([
  happyPackBlock(babel(), { ...happyRelatedOptions }),
])
```

We'd also need to take into account cross-loader options (like sharing a thread-pool between different HappyPack plugin instances) which I've pointed to in the `...happyRelatedOptions` part. The block could make the necessary adjustments to the config such that we end up with the right plugin definitions and so on.

andywer:
@amireh I think the main goal to focus on right now is to create a webpack block for happypack that works and is easy to use. We can then add it as package under the `@webpack-blocks` package scope :)

Unfortunately I on the other hand have got no experience with happypack 😉 But we could still try to team up here. I can improve the documentation on how to write webpack-blocks and answer your questions and you can try to contribute a block for happypack.

happypack's API really makes it rather hard to use / write a block for it (kinda like [extract-text](https://github.com/andywer/webpack-blocks/tree/master/packages/extract-text)). But the good thing is that once it's done, it will be way easier to get happypack up and running :)

Btw: It's pretty new, so it's no surprise you haven't heard of it yet ^^

fenos:
@amireh @andywer Is there any progress on this? I'm very glad of using `webpack-blocks` for my configs, but I couldn't find a way to build working a block with `happypack`, is something blocking this?

andywer:
[...]

PR #126, opened 2017-03-01
Happypack block

fenos:
As discussed in the issue #14 will be very nice to have a `happypack` block wich allows us to easily implement `Happypack` features to improve the performance of the build.

@andywer found a neat way to implement the block using the `webpack-block` API, big thanks for that. 

With a bit of delay I finish it up, so here you go.

```js
  happypack([
     babel(),
     sass(),
  ])
])
```

**Note:**
I'm using the loader for a while now, and unfortunately, I haven't seen any improvements in terms of build speed in either **watch mode** and **production build** This Pr will also help to debug the problem.

### DIFF
--- a/packages/happypack/index.js
+++ b/packages/happypack/index.js
@@ -62,7 +62,7 @@ function happyfyConfig(configSnippet, _, happypackConfig) {
  * Takes an array, a function or something falsy and returns the array, the
  * function wrapped in an array or an empty array, respectively.
  */
-function toArray(value) {https://github.com/fenos/webpack-blocks
+function toArray(value) {
   if (value) {
     return Array.isArray(value) ? value : [value]
   } else {
--- a/packages/happypack/index.js
+++ b/packages/happypack/index.js
@@ -1,0 +1,71 @@
+const { group } = require('@webpack-blocks/core')
+const HappyPack = require('happypack');
+const os = require('os');
+
+module.exports = happypack
+
+function happypack(blocks, configs = {}) {
+  return group(blocks.map((block) => happyfyBlock(block, configs)))
+}
+
+/**
+ * Returns a new block wrapping `block` that creates a happypack loader config.
+ */
+function happyfyBlock(block, config) {
+  const happyBlock = happyfySetter(block, config)
+  const pre = toArray(block.pre)
+  const post = toArray(block.post).map(postHook => happyfySetter(postHook, config))
+
+  return Object.assign(happyBlock, { pre, post })
+}
+
+/**
+ * Takes a block or a post hook and returns a wrapping function that creates a happypack loader config.
+ */
+function happyfySetter(setter, happypackConfig) {
+  return (context, config) => happyfyConfig(setter(context, config), context, happypackConfig)
+}
+
+/**
+ * Transforms a non-happypack loader config into a happypack loader config.
+ */
+function happyfyConfig(configSnippet, _, happypackConfig) {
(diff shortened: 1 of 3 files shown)


packages/happypack/index.js
```diff
@@ -0,0 +1,71 @@
...
+    }
+  }
+  return {};
+}
+
+/**
+ * Takes an array, a function or something falsy and returns the array, the
+ * function wrapped in an array or an empty array, respectively.
+ */
+function toArray(value) {https://github.com/fenos/webpack-blocks
```
tb: https://github.com/fenos/webpack-blocks ??

andywer:
@tb Thanks :) And let's thank @github for letting the target repo owner edit files of PRs 😉

andywer:
@amilajack We still need some proof that it works as expected, esp. due to the [performance concern](https://github.com/andywer/webpack-blocks/issues/14#issuecomment-281289489). Can we do anything to help testing it?

amilajack:
@andywer What kind of testing would you like me to do? I can test this on some projects that I have that use webpack. Or do you want me to do performance testing?

andywer:
Just some rudimentary testing. Does it work (I suppose so) and is it faster than without :)

fenos:
@amilajack @andywer I'm preparing a small react repo with webpack-blocks so we can measure the performance

@andywer : thanks for fixing the copy typo

fenos:
@andywer @amilajack I was trying to find a react starter trough github so that i could just swap the webpack config to webpack-block, they are all relatively small where the build takes less then 4.0sec, so is not realistic. However there I could see that happypack did improve performance of 1s. 

So I'm wondering if there is something wrong with the webpack config that we are using, or maybe not. 

If you could try it in a bigger project then would help

amilajack:
@fenos A relatively big project that I would recommend is https://github.com/chentsulin/electron-react-boilerplate, which I am a maintainer of. In the project, we're having a lot of issues with slow build times. Also to have noticeable performance improvements, I think we have to add support for webpack caches and DLL's

andywer:
Any news?

amilajack:
Haven't had the chance to do this yet. I hope @fenos had a go at this

pigcan:
Any news?

diegohaz:
Hey, guys. FYI, I've just published [`webpack-blocks-happypack`](https://github.com/diegohaz/webpack-blocks-happypack) based on this PR.

It has a `DISABLE_HAPPY` env flag so we can easily test if it's actually improving build speed.

On my tests, I'm using it on [ARc](https://github.com/diegohaz/arc), which is a relatively small project, and these are the build times on my Mac (4 CPUs):
- without happy: ~11.5s;
- with happy and without cache: ~10.5s;
- with happy and cache: ~9.5s.

ARc only uses `babel-loader` though.

I'm also using that on a bigger project at work with `babel-loader` and `css-loader`. Time decreased from ~40s to ~30s.

If this PR get merged, I can deprecate that. I'm also ok maintaining it as I'm using it on big projects in production.

andywer:
Hey @diegohaz. Sounds great, thank you!

We should look into that, esp. since it's already published.
And I promise I will come up with a community blocks list soon to keep track of all the awesome stuff that's not in the main repo 😉

sapegin:
So what we’re going to do here? I’d close this PR, make a list of third-party blocks (already planned and needed) and link @diegohaz block there.

andywer:
I agree with @sapegin

PR #128, opened 2017-03-02
New API

### LINKED ISSUE #125: API for version 1.0
After thinking a lot about it quite a long time, encountering several edge-cases with the current API and a long discussion with @aaronjensen I want to present the draft for a slight API change.

It is a breaking change, but migration is straight-forward. The version bump 0.x => 1.0/2.0 seems to be an obvious choice for introducing such a change.

## Old (current) API

```js
type Block = (context, prevConfig) => configSnippet
```

## Block written for old (current) API

```js
module.exports = function exampleBlock () {
  return (context, prevConfig) => ({
    module: {
      loaders: [
        {
          test: context.fileType('text/css'),
          loaders: ['sample-css-loader']
        }
      ]
    }
  })
}
```

The main issues here are:
* Implicit merge of the returned snippet
* No control over how this config snippet will be merged

(Pre and post hooks work as before, just that the post hooks will adapt the new API as well. Pre hooks don't actually emit configuration, so they don't need to be changed.)

## New API

```js
type Block = (context, helperFunctions) => prevConfig => nextConfig
```

## Block written for new API

```js
// With `merge()` helper:

module.exports = function exampleBlock () {
  return (context, { merge }) => merge({
    module: {
      rules: [
        {
          test: context.fileType('text/css'),
          use: ['sample-css-loader']
        }
      ]
    }
  })
}

// With `addLoader()` helper:

module.exports = function exampleBlock () {
  return (context, { addLoader }) => addLoader({
    test: context.fileType('text/css'),
    use: ['sample-css-loader']
  })
}

// Without any helpers:

module.exports = function exampleBlock () {
  return context => config => ({
    ...config,
    module: {
      ...config.module,
      rules: config.module.rules.concat([
        {
          test: context.fileType('text/css'),
          use: ['sample-css-loader']
        }
      ])
    }
  })
}
```

With the new API the modification of the configuration object is not just more explicit, but the main feature is that the block now has full control on *how* to merge new content into the configuration, like if some item will be prepended or appended to an array.

## List of helper functions (Might still be subject to change)

- `merge(configSnippet: object)`
- `addLoader(loaderDefinition: object)`
- `addPlugin(pluginInstance: Plugin)`

@eXon @jvanbruegge @sapegin

andywer:
Introduces the new core API. Resolves #125.

Still using `module.loaders` and webpack-merge's smart merge, though.

@aaronjensen


andywer:
@jvanbruegge @eXon Feel free to review as well 😉

docs/BLOCK-CREATION.md
```diff
@@ -38,6 +40,50 @@ function babel (options) {
...
+```js
+function sampleBlock () {
+  return (context, { merge }) => merge({
+    resolve: {
+      extensions: ['ts']
+    }
+  })
+}
+```
+
```
aaronjensen: I'd add `group` to this list

packages/babel6/__tests__/babel6.test.js
```diff
@@ -1,15 +1,20 @@
...
 
+  const addLoader = sinon.spy((loaderDef) => config => config)
   const context = {
     fileType: () => '*.js'
   }
 
-  t.deepEqual(block(context), {})
-  t.deepEqual(block.post(context).module.loaders, [
+  t.deepEqual(block(context, { addLoader })({}), {})
+  t.deepEqual(block.post(context, { addLoader })({}), {})
```
aaronjensen: addLoader is a pure function, so it's usually a better idea to include/use the real one and do state based testing rather than setting it up as a spy. In general, I'd consider pushing towards doing snapshot testing for blocks, but in the meantime at least using the real utility functions would make these tests more valuable imo.

Also, I'm not entirely convinced that unit testing blocks is worthwhile. You get more bang for the buck by actually testing them in createConfig, I think.

packages/css-modules/index.js
```diff
@@ -23,18 +23,14 @@ function cssModules (options) {
...
-          loaders: [ 'style-loader', 'css-loader?' + stringifyQueryParams({ importLoaders, localIdentName, modules: true }) ]
-        }, options.exclude ? {
-          exclude: Array.isArray(options.exclude) ? options.exclude : [ options.exclude ]
-        } : {})
-      ]
-    }
-  })
+  return (context, util) => util.addLoader(
+    Object.assign({
+      test: context.fileType('text/css'),
```
jvanbruegge: Does this work with sass too?

packages/core/package.json
```diff
@@ -18,20 +18,43 @@
     "functional"
   ],
   "ava": {
+    "babel": {
+      "plugins": [
+        [
+          "transform-object-rest-spread",
```
aaronjensen: I thought ava included this by default? https://github.com/avajs/ava#es2017-support

packages/typescript/index.js
```diff
@@ -44,6 +42,7 @@ function typescript (options) {
 
 function pre (context) {
   const registeredTypes = context.fileType.all()
+
   if (!('application/x-typescript' in registeredTypes)) {
```
jvanbruegge: Would remove this pre hook, as it is now in the default mime types and the 2.0 update is a breaking one, so you have to update

packages/core/src/index.js
```diff
@@ -94,11 +93,20 @@ function getHooks (configSetters, type) {
...
 }
 
-function invokeConfigSetters (configSetters, context, baseConfig = {}, getCompleteConfig = (config) => config) {
+function invokeConfigSetters (configSetters, context, baseConfig) {
   return configSetters.reduce(
-    (mergedConfig, setter) => {
-      const configPartial = setter(context, getCompleteConfig(mergedConfig))
-      return merge.smart(mergedConfig, configPartial)
+    (config, setter) => {
+      const updateFunction = setter(context, helpers)
```
aaronjensen: helpers -> util

packages/extract-text/index.js
```diff
@@ -19,26 +19,28 @@ function extractText (outputFilePattern, fileType) {
...
+        loader: nonStyleLoaders
+      })
+    })
+    const _addPlugin = util.addPlugin(plugin)
+    const _removeLoaders = removeLoaders(context.fileType(fileType))
+
+    // Now apply them to the config: Remove the existing loaders for that
+    // file type, add the new loader and add the plugin
+
+    return _addPlugin(_addLoader(_removeLoaders(prevConfig)))
```
aaronjensen: good place to use `group`

packages/css-modules/index.js
andywer: Don't think so, but never tried it to be honest... Good point 🤔

packages/typescript/index.js
andywer: True

packages/babel6/__tests__/babel6.test.js
andywer: Yeah, I have got mixed feelings about unit testing the blocks, too. Integration tests are definitely worth more. But I didn't want to open another construction site right now...

The reason why I didn't use the real helpers was just that the unit test would then not only test this unit, but also the utils. Also using a spy makes testing quite easy.

But yeah, we should consider turning the block unit tests into integration tests.

packages/core/package.json
andywer: I thought so, too. But it seems quite a lot of features that once were part of the es2017 preset have been removed from it recently. Object-spread isn't included anymore... :-/

https://babeljs.io/docs/plugins/preset-es2017/

packages/extract-text/index.js
andywer: But it isn't a group of blocks, it's just three update functions... I already thought about `util.aggregate([ _removeLoaders, ... ])` or something like that, but I'm hesitating adding more new stuff to `util`.

packages/babel6/__tests__/babel6.test.js
aaronjensen: > The reason why I didn't use the real helpers was just that the unit test would then not only test this unit, but also the utils. Also using a spy makes testing quite easy.

Right, but it's misusing spies imo. Spies should only be used for side effecting things. These are pure functions. It's ok to "test" them, because what you're really testing is that they do what you think they're doing in this specific context. I wouldn't even consider it an integration test. You shouldn't have things that return values be spies, it's like replacing `merge` with a spy, or `mapValues`. 

I don't think there's anything in webpack-blocks that warrants testing w/ sinon, there are no side effects. It's all pure functions which is part of its beauty.

packages/core/package.json
aaronjensen: Ah, indeed, it's in experimental.

packages/css-modules/index.js
jvanbruegge: If not, use a post hook to query all loaders with style loader and use that mime type

packages/extract-text/index.js
aaronjensen: I mentioned `flow` before, it's a lodash function. It's what you need here (though you probably don't need to pull it in from lodash unless you want to). And I'd say it's a must have. It's a key part of function composition which is what this api *should* drive people to.

I prefer [my implementation](https://github.com/substantial/webpack-parts/blob/d17ab43e887b4b7104b8c708bf10c0d07835107c/index.js#L67-L68) of `flow` because it flattens and removes nils allowing you to compose like this:

```
flow(
  option && doThing(),
  option2 && [doThing1(), doThing2],
  doOtherThing(),
)
```

packages/babel6/__tests__/babel6.test.js
aaronjensen: Btw, this is a good video about spies vs mocks vs not using them https://www.youtube.com/watch?v=URSWYvyc42M

andywer: > I wouldn't even consider it an integration test.

It's not an integration test, right now it's a unit test. But maybe we talk past each other here. It will probably be best to turn it into an integration test, using it in `createConfig()` and look at the resulting config.

Agree? Might do it in another PR, though.

aaronjensen: Right, and it would still be a unit test even if it used the real util functions. That was my point. I'd urge you to remove sinon as a dependency and do that instead (another pull is fine) or port to a test that uses `createConfig` (and still remove sinon). I don't think sinon is a good fit for functional projects and I wouldn't want to encourage contributors to misuse it.

packages/extract-text/index.js
andywer: Hmm, that might be a candidate to be included in `util`...

packages/css-modules/index.js
andywer: I opened a follow-up issue for this: #130.

andywer:
So as far as I can see there are no immediate issues left to fix for this PR.

One topic that might be considered part of this PR is:
*Should we add a `flow()` helper as pointed out by @aaronjensen?*

One topic that relates to this PR, but can be settled later:
*Should the built-in utility functions be dropped and moved to a separate package? (#125)*

andywer:
Merged it, since the open topics don't block this PR.

PR #129, opened 2017-03-05
Add NamedModulesPlugin when HotReloading is active

bkonkle:
This is *very* handy for debugging. 😀


andywer:
Very good point! Thanks, @bkonkle :)

Just one concern: This is webpack 2 only, right?

andywer:
Poking 👉 @bkonkle

sapegin:
@andywer Should we merge this? 1.0 will work only with webpack 2 anyway.

andywer:
Good point, @sapegin. Will rebase it on the 1.0 release branch later today.

PR #135, opened 2017-04-01
webpack-common: hotOnly by default for devServer

langovoi:
Hello :-)

I think that `webpack-blocks` don't need to worry about `webpack/hot/only-dev-server`, because users of `webpack-blocks` can self use `hotOnly` and `hot` options of `webpack-dev-server` (for example switch off it)

I setted options `hotOnly` by default for backwards compatibility (`hot` been setted previously)

### DIFF
--- a/packages/webpack-common/src/__tests__/devServer.integration.test.js
+++ b/packages/webpack-common/src/__tests__/devServer.integration.test.js
@@ -22,6 +22,7 @@ test('devServer() without options provides expected defaults', (t) => {
     },
     devServer: {
       hot: true,
+      hotOnly: true,
       historyApiFallback: true,
       inline: true
     },
@@ -51,6 +52,7 @@ test('devServer() uses custom options and can be composed', (t) => {
     },
     devServer: {
       hot: true,
+      hotOnly: true,
       historyApiFallback: true,
       inline: false
     },
--- a/packages/webpack-common/src/__tests__/devServer.integration.test.js
+++ b/packages/webpack-common/src/__tests__/devServer.integration.test.js
@@ -18,7 +18,7 @@ test('devServer() without options provides expected defaults', (t) => {
 
   t.deepEqual(config, {
     entry: {
-      main: ['./test.js', 'webpack/hot/only-dev-server']
+      main: ['./test.js']
     },
     devServer: {
       hot: true,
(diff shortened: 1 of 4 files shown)


andywer:
Hey @langovoi.

Very true! Thanks for pointing out.
I guess this is a remnant of the very early days.

Will happily merge it, but you might have to wait for the 1.0 release to see the fix in action. Or do you need it urgently?

langovoi:
@andywer glad to see that you like it.

No, it not need urgently, but I think will be better to release this hotfix in the near future.

andywer:
Did it break something in your use case?

langovoi:
It's break when I use devServer with hotOnly option:
```javascript
createConfig([
  ...,
  devServer({
    hotOnly: true
  })
])
```

```
Invalid configuration object. Webpack has been initialised using a configuration object that does not match the API schema.
 - configuration.entry should be one of these:
   object { <key>: non-empty string | [non-empty string] } | non-empty string | [non-empty string] | function
   The entry point(s) of the compilation.
   Details:
    * configuration.entry['accounts'] should be a string.
    * configuration.entry['accounts'] should NOT have duplicate items (items ## 1 and 4 are identical) ({
        "keyword": "uniqueItems",
        "dataPath": ".entry['accounts']",
        "schemaPath": "#/definitions/common.nonEmptyArrayOfUniqueStringValues/uniqueItems",
        "params": {
          "i": 4,
          "j": 1
        },
        "message": "should NOT have duplicate items (items ## 1 and 4 are identical)",
        ...
        "data": [
          "/Users/marklangovoi/WebstormProjects/accounts/node_modules/webpack-dev-server/client/index.js?http://0.0.0.0:8080",
          "webpack/hot/only-dev-server",
          "react-hot-loader/patch",
          "/Users/marklangovoi/WebstormProjects/accounts/accounts/index.jsx",
          "webpack/hot/only-dev-server"
        ]
      }).
```

Also i think only one of ```webpack/hot/dev-server``` and ```webpack/hot/only-dev-server``` must be included to entries points.

- ```hot``` option add ```webpack/hot/dev-server```
- ```hotOnly``` option add ```webpack/hot/only-dev-server``` (```webpack/hot/dev-server``` will be not added, if ```hot``` enabled)

https://github.com/webpack/webpack-dev-server/blob/85de417fd5592efc5448758bf550d7d34e670f60/lib/util/addDevServerEntrypoints.js#L9

langovoi:
What about to release this fix?

andywer:
It can be released as it is, but need to release a new 0.x version. Would like to publish it together with more changes, but there don't seem to be any right now.

So we might just publish a mini-release for this single fix... Can do it tomorrow, I think.

PR #138, opened 2017-04-04
webpack must be written in lowercase

langovoi:
Hello ;-)

webpack must be written in lowercase. About It written in [webpack/media](https://github.com/webpack/media#name)


andywer:
Ahh, really? Thanks for pointing out :)

I could swear I've seen it written with leading upper-case W on the official webpack site...

But in the `package.json` description it just looks wrong :-/
```
"description": "webpack block for CSS modules."
```

What do we do about it? Maybe restructure the sentence like "The CSS modules webpack block."?

sapegin:
Prefer common sense over weird rules: webpack should be capitalized in the beginning of the sentence just to make users‘ lives easier.

andywer:
I agree. It will probably be hard to consistently stick to it across different developers, if the rule contradicts intuition too much.

Btw, I just resolved the merge conflict that occured due to the postcss block version bump. I didn't touch the changed `description`, though, when resolving the `package.json` conflict.

andywer:
So... How about lower-case webpack everywhere, but not at the beginning of a sentence?
Because it looks weird and people will start to violate it, I am almost certain. And I don't feel like paying particular attention to it.

I am confident that the webpack boys won't set us on fire for this decision 😉

sapegin:
That’s what I [use myself](https://github.com/sapegin/textlint-rule-terminology/blob/master/terms.json#L40-L43) ;-)

andywer:
What does @langovoi say? 😉

langovoi:
Okey guys! If writing of webpack in begin of sentence with a capital letter will be more clearly, I will fix it.

Hope guys from webpack don't be offended 😌

andywer:
FYI The lower-case w in webpack isn't written in stone ^^ https://twitter.com/wsokra/status/855800490713649152

andywer:
If we ignore the upper-cased webpack at the beginning of sentences / headlines it boils down to changing a few lines in the main README, right?

We could reduce the PR to this for now and maybe look back at this thing later when/if they decide to update the convention. Would you be ok with that? 😉

andywer:
Closing this PR for now, since it's outdated, non-critical and would require further changes before merging.
Feel free to re-open, though 😉

PR #140, opened 2017-04-05
Update env function, to support more cli parameters

ArsSirek:
Hi all, thanks for the great tool!)

I feel like my case isn't much relatable, but would like to hear your feedback.
We have env function from core, and it allows us to make some parts conditional, based on NODE_ENV, but we can't share some conditional configuration among environments, can we?
For example, I wanted to add _webpack-bundle-analyzer_ to dev or build configuration, when that is explicetly required from cli... I feel that this is not right thing to do, but we can addres a question in comman.

Based on [this comment](https://github.com/webpack/webpack/issues/2254#issuecomment-203528744)
we can access custom cli parameters nativity with webpack 

If we wrap createConfig call in function that  receives cli env as argument, we can add conditions like this

`NODE_ENV=production webpack --config webpack.config.babel.js --env.analyse`

```

module.exports = (environment) => {
    environment = environment || {};
    let conditionalSetters = [];
    if (environment.analyse) {
        conditionalSetters.push(
            // partly common setters
        );
    }
    return createConfig([
        //regular config
        ...conditionalSetters
    ]);
}
```
I wasn't able to find a way to bring this arguments to env function to extent its logic, is there any? or is this a good way to go about it? 

```
module.exports = (environment) =>  createConfig([
    //regular config
    groupIf(environment.analyse,conditionalSetters)
]);

```

With best regards

### DIFF
--- a/packages/core/src/index.js
+++ b/packages/core/src/index.js
@@ -59,7 +59,7 @@ function env (envName, configSetters) {
         }
     }
 
-    if (addSettings) {
+    if (!addSettings) {
         return () => ({})
     } else {
         return group(configSetters)
--- a/packages/core/src/index.js
+++ b/packages/core/src/index.js
@@ -50,14 +50,14 @@ function createConfig (initialContext, configSetters) {
 function env (envName, configSetters) {
   const currentEnv = process.env.NODE_ENV || 'development'
 
-  return groupIf(currentEnv !== envName, configSetters);
+  return groupIf(currentEnv === envName, configSetters);
 }
 
 /**
  * Applies an array of webpack blocks only if condition is true
  */
 function groupIf(condition,configSetters) {
-    if (condition) {
+    if (!condition) {
         return () => ({})
     } else {
         return group(configSetters)
--- a/packages/core/src/index.js
+++ b/packages/core/src/index.js
@@ -5,6 +5,7 @@ const defaultFileTypes = require('./defaultFileTypes')
 exports.createConfig = createConfig
 exports.group = group
 exports.env = env
+exports.groupIf = groupIf
 
 const isFunction = (value) => typeof value === 'function'
 
@@ -49,11 +50,18 @@ function createConfig (initialContext, configSetters) {
 function env (envName, configSetters) {
   const currentEnv = process.env.NODE_ENV || 'development'
 
-  if (currentEnv !== envName) {
-    return () => ({})
-  } else {
-    return group(configSetters)
-  }
+  return groupIf(currentEnv !== envName, configSetters);
+}
+
+/**
+ * Applies an array of webpack blocks only if condition is true
+ */
+function groupIf(condition,configSetters) {
+    if (condition) {
+        return () => ({})
+    } else {
+        return group(configSetters)
+    }
 }
 
 /**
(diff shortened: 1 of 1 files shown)


andywer:
Hi @ArsSirek and thanks.

I think the PR's changes do not really correlate with what you are writing. Do I get you right? You want `env()` to be able to not only compare `process.env.NODE_ENV` but any `process.env.*` value (like `process.env.analyse`)?

Maybe we could make env() accept an additional parameter or an object for this...
👉 `env({ analyse: true }, [ ... ])`

ArsSirek:
Hi @andywer ! and thanks for the quick response,
Yea, except I didn't really know that we can access it like you showed!

This pull request was more to show the idea. So, we don't need to use webpack features to access arguments passed with cli command, right? Then, I think, that adding some syntax to be able to check conditions based on other cli parameters would be useful.

Your suggestion with option to pass object as 1 parameter looks cool to me.
I will try to change the request, to show if I get it right this time.

andywer:
Sorry for the delayed response!
I think I misunderstood the new --env behavior. Had another look at it and now I understand your point better :)

`--env` does not alter `process.env` as far as I can see, so my suggestion was not very useful.

Your suggestion is very valid, sorry for the confusion :)
I am not completely happy with new name `groupIf` yet, although it describes its quite well. Will think about it some more...

ArsSirek:
Hi @andywer, that's not even delayed, and this is just a discussion anyway, but thanks!
I actually played with cli to confirm that too)
You are right, "groupIf" name is not good. ( I didn't like it either, just hadn't better ideas)
 In fact, I think, adding "groupIf" doesn't seem to be worth attention from users, because, the same could be achieved with inline function..
I originally was just thinking if it is possible (worth), to support  "the new --env behavior" in webpack-blocks. 
 But, after your suggestion, I get that we can use node environment params with process.env here. And I don't know why one would choose to use --env.param in weppack call, instead of node environment param. I just don't know details with node deep enough.
 I updated env function to show, how I get to use suggested syntax. Not sure if it is fine, because we can't actually use Boolean in cli, so I used not strict check, and that may not be too obvious. 

My use case is very small, and I can go with inline functions, so, how do you think, will this feature be useful? should we continue the work? I am fine with any decision.

andywer:
I think if we add anything then maybe something to make that `someCondition ? group([ ... ]) : group([])` easier. But I am not yet sure if people actually need that.

I guess the `--env` parameter was just added to webpack, since before webpack 2.0 you could pass any parameters to the webpack cli, it would not complain and loaders/plugins would check the parameters for stuff related to them. But since webpack 2 throws an error on unknown parameters, this is probably the work-around 😉

Suggestion: Let's drop the PR for now, but remember the idea. If the conditional block group thing pops up again we can have another look :)

andywer:
@ArsSirek Check out https://www.npmjs.com/package/webpack-blocks-utils#envvarkey-value-configsetters by @ecliptic/@bkonkle. This seems to be pretty much what we talked about :)

ArsSirek:
Hi @andywer, yeah, thanks for the link, we are not alone). That [envVar function](https://github.com/ecliptic/webpack-blocks-utils/blob/master/lib/index.js#L40) code is more readable, but it's a separate function with different name. 

  If would ask me, I like [overloading](https://en.wikipedia.org/wiki/Operator_overloading) approach more, even though for JavaScript it ends up with ugly type checks ) But I am not experienced enough to state benefits, it's more like taste question to me.

  Sorry, I didn't response to the previous comment. I absolutely agree with suggestion to close. I too don't like the code as it is now. I am sure we will found better ways when we will see the need. )

Have a great day!

Issue #91, opened 2017-01-18
Add testing/setup documentation for contributors


andywer:
## Package linking / testing before publishing

- `lerna bootstrap`
- Point out that `lerna bootstrap` is run automatically on `postinstall` (on Travis CI, too)
- `npm link`

## Testing concept

Types of tests:
- Unit tests
- Integration tests
- End-to-end tests

(What they reside, where they occur and when to write which one(s))

## Common pitfalls

- e2e tests: Copy `dependencies` of blocks to test into the `devDependencies` of `webpack`/`webpack2`

sapegin:
We could close this soon :-) 

* Package linking / testing before publishing — will be done in #185
* Testing concept — done already
* Common pitfalls — not sure about that

sapegin:
> Common pitfalls — not sure about that

What can we do to solve this?

andywer:
Yeah, let's close it for now 😉

PR #141, opened 2017-04-10
Pass typescript options via query string

jvanbruegge:
This allows you to pass options not only via tsconfig.json but also via webpack.config.js
I would need this for https://github.com/cyclejs-community/create-cycle-app/issues/103


jvanbruegge:
I tested it locally and it works as expected

andywer:
Good point. And as always thanks for the PR :)

But `tsconfig` and `compiler` options will now be passed as JSON options and via `TsConfigPathsPlugin`. Maybe it doesn't do any harm, but doesn't seem completely right either. Maybe separate the TsConfigPathsPlugin options from the loader options in the beginning?

jvanbruegge:
I missunderstood the plugin, this is better now

andywer:
Ok, now I am confused ^^
I can't find any `advancedPath` option on https://github.com/s-panferov/awesome-typescript-loader#loader-options

jvanbruegge:
yes, because normally you would either add the plugin or not, the option just handles inclusion

andywer:
Hmm, I like the proposed change in general, but I don't like having an additional custom option for utility purposes only. Maybe `options.paths` for `tsconfig` and `compiler`, the rest of `options` becomes loader options.

What do you say?

jvanbruegge:
the loader needs the compiler and tsconfig too

jvanbruegge:
i never used this feature (and i guess noone else too), so this never surfaced

andywer:
But what is the `TsConfigPathsPlugin` good for if we have to pass the paths to the loader anyway?

Just recognized you were the autor of the block, so in case of doubt I tend to trust what you say 😉
But it still seems a little strange.

jvanbruegge:
Im not completely sure either, the docs dont say why

andywer:
Maybe it's worth opening an issue at the awesome-typescript-loader's repo?

jvanbruegge:
yes i can do that.

jvanbruegge:
There is already one, but it seems I use it right now

andywer:
How about dropping the `options.advancedPaths` and go back to the `options.tsconfig || options.compiler` condition for the `TsConfigPathsPlugin` if we need it plus the complete JSON config?

(Especially since `@param {boolean} [options.advancedPath] See https://github.com/s-panferov/awesome-typescript-loader#loader-options` seems misleading and the documentation for `compiler` & `tsconfig` has been dropped)

The rest looks fine to me :)

jvanbruegge:
I finally understood what the plugin is for. We can have it active all the time. It only affects people configuring it in the tsconfig.json. Thats why it needs the path. I will remove the check

jvanbruegge:
Should I remove the unit test? It's actually quite pointless

andywer:
Nice! Yes, I think it's safe to remove the unit tests. End-to-end tests are way more reliable for testing blocks. Poking #91...

But adding a comment explaining why we need the `TsConfigPathsPlugin` and why it's fine to feed it with `null`/`undefined` would be helpful.

jvanbruegge:
Can we merge this?

andywer:
Of course 👌

andywer:
Published as `@webpack-blocks/typescript v0.4.1`

PR #144, opened 2017-04-19
Fix tsconfig option name

jvanbruegge:
The option is called `configFileName` for the loader, but `tsconfig` for the plugin


andywer:
Damn it 😉

PR #146, opened 2017-04-26
1.0 release: Drop support for node < 6

andywer:
Allows usage of default parameters, destructuring and some more important syntactic features without transpiling.


packages/elm/index.js
```diff
@@ -14,9 +14,10 @@ module.exports = elm
  * @param {string}                  [options.cwd]           Custom location for your elm files.
  * @return {Function}
  */
-function elm (options, isProduction) {
-  options = options || {}
-  isProduction = isProduction || process.env.NODE_ENV === 'production'
+function elm (options = {}, isProduction) {
```
jvanbruegge: Why this boolean at all and not directly `NODE_ENV`?

andywer: I didn't write the block and I agree, a more elegant solution than a isProduction flag would be nice. (The bad boy that I am) I fixed a small issue I recognized on-the-fly while doing the node 6+ thing, even though it's not related.

But I think we can live with it for now. Not sure how many people currently use the block. I would say let's improve it once we know it is really used?

andywer: Btw, what exactly did you mean by "and not directly NODE_ENV"?

jvanbruegge: `NODE_ENV === 'production'`

jvanbruegge: as replacement for the boolean.

andywer: I think the idea was "let the user decide if he needs a development or production config, but use a sane default".

Suggestion: Let's open an issue for that if you think it is a problem.

andywer: But yeah... As I think about it. Since it's just for selecting the defaults: Maybe it would be best to just use `NODE_ENV === 'production'` for selecting the defaults, without parameter.

PR #147, opened 2017-04-26
Stop spurious usage warnings in console

SteveALee:
Without this fix the following warnings appear in the console:

fallbackLoader option has been deprecated - replace with "fallback"
loader option has been deprecated - replace with "use"

### DIFF
--- a/packages/extract-text2/index.js
+++ b/packages/extract-text2/index.js
@@ -31,8 +31,8 @@ function extractText (outputFilePattern, fileType) {
             test: context.fileType(fileType),
             exclude: loaderConfig.exclude,
             loader: plugin.extract({
-              fallbackLoader: 'style-loader',
-              loader: nonStyleLoaders
+              fallback: 'style-loader',
+              use: nonStyleLoaders
             }),
             loaders: undefined
           }


packages/extract-text2/index.js
```diff
@@ -31,8 +31,8 @@ function extractText (outputFilePattern, fileType) {
             test: context.fileType(fileType),
             exclude: loaderConfig.exclude,
             loader: plugin.extract({
-              fallbackLoader: 'style-loader',
-              loader: nonStyleLoaders
+              fallback: 'style-loader',
+              use: nonStyleLoaders
             }),
```
SteveALee: This might break in older versions of webpack 2

jvanbruegge: We are currently working on webpack-blocks version 1.0, that will fix this completely.
You can disable the warning with `process.noDeprecations = true`

SteveALee: OK. Feel free to reject then as this is a nit really
Where does 'process.noDeprecations = true' go?

@jvanbruegge - you seem to be everywhere :)

jvanbruegge: just inside your webpack config

jvanbruegge: You can see it here: https://github.com/cyclejs-community/create-cycle-app/blob/39296f083f93cafbb31afbf928c7965ed0329228/packages/cycle-scripts/configs/webpack.config.template.js#L4

SteveALee: Tah - and you figured the context too :-P

SteveALee:
Closing as WIP will fix

andywer:
I will happily move the patch to the 1.0 release branch, though :)

PR #148, opened 2017-04-27
1.0 Feature: Asset loaders

andywer:
Drops the implicit default configuration that came with `createConfig()` and replaces it with a new `assets` package that contains a `css`, `css.modules`, `file` and `url` block.

Also removes the hacky `createConfig.vanilla()`.

@sapegin @jvanbruegge @eXon @aaronjensen @Veikedo


packages/assets/lib/css.js
```diff
@@ -0,0 +1,82 @@
...
+ * @return {Function}
+ * @see https://github.com/webpack-contrib/css-loader
+ */
+function css (fileType = 'text/css', options) {
+  if (fileType && typeof fileType === 'object' && !options) {
+    options = fileType
+    fileType = 'text/css'
+  }
+  if (!fileType || typeof fileType !== 'string') {
+    throw new Error(`Need to pass a valid file type to css().`)
```
sapegin: Something like “File type should be string” would be more clear.

jvanbruegge:
nothing to add to @sapegin's review

andywer:
Good points. Thanks for the review!

Tell me if it's fine for you or if you think there are still points to discuss.

sapegin:
Definitely better!

Issue #134, opened 2017-03-15
1.0 release


andywer:
Hi everyone! This is gonna be the "1.0 release issue", showing the big picture, in contrast to the many small issues.

## Version number change

Since webpack 1.x has [officially been deprecated](https://webpack.github.io/docs/) I am now fine with releasing the next webpack-2-only major version as `webpack-blocks 1.0`. So if you find references to a 2.0 release, this will actually be the 1.0 release.

## Changes

- [x] Webpack 2 only, drop duplicate packages (like `extract-text` and `extract-text2`) #127
- [x] Use `module.rules` syntax #111 #124
- [x] Implement new API: #125
- [ ] Decide if the new API and its `util` parameter are fine or should be changed #125
- [ ] Drop `release/1.0` branch, rename `release/2.0` => `release/1.0`
- [ ] Make `createConfig.vanilla()` the new `createConfig()`
- [ ] Create a new package with CSS, font, image, ... loaders (because of the change above)
- [ ] Make sure that different CSS loaders work seemlessly with `node_modules` exclusion #133
- [ ] Improve sass + extract-text documentation #115
- [ ] Merge #117
- [ ] Rebase on master

(Not necessarily in that order 😉)

## Optionally

- [ ] Check which parts of `webpack-parts` might be merged into webpack-blocks #122
- [ ] Check if SASS works together with CSS modules #130

@eXon @jvanbruegge @sapegin @aaronjensen

andywer:
Sorry for shooting ahead, but since we already had the discussion about 1.0/2.0 I figured there is not much sense in riding the now-deprecated horse much further.

You guys agree with the short summary of the 1.0 release?

jvanbruegge:
I would add that extract text should use a post hook to work automaticly with sass (so you dont need to document anything). Would make order of sassBlock and extract-text block irrelevant

andywer:
Good point!

aaronjensen:
This looks good to me. I probably wouldn't rebase onto master at this point (too much risk, merge is safer and maintains history) but that's a workflow nit. I'm very 👍  on webpack 2 only.

andywer:
True! Updated the list :)

andywer:
Update: "Make createConfig.vanilla() the new createConfig()" and "Create a new package with CSS, font, image, ... loaders" are almost done 🎉

[...]

PR #149, opened 2017-05-07
webpack-blocks convenience package (for 1.0)

andywer:
Introduces the "webpack-blocks" convenience package that wraps the following packages:

* @webpack-blocks/assets
* @webpack-blocks/babel6
* @webpack-blocks/dev-server
* @webpack-blocks/extract-text
* @webpack-blocks/postcss
* @webpack-blocks/typescript
* @webpack-blocks/webpack

Packages in the repo not contained in this package:

* @webpack-blocks/elm
* @webpack-blocks/sass
* @webpack-blocks/tslint

Please tell if you are ok with the included packages or not! 😉
I am especially unsure about the `sass` package, since it's frequently used, but also takes some noticeable time to install.

Also:
Bumped all package versions to `v1.0.0-alpha`, so we can publish an alpha release.

@jvanbruegge @eXon @sapegin #134


sapegin:
I’d extract TypeScript and PostCSS too for consistency.

jvanbruegge:
I would include them as they are used very frequently. And unlike sass, they install fast

andywer:
@sapegin Can you explain your thought a little further?

I think PostCSS is important, since everyone using css or css modules will want to use at least the autoprefixer, I guess.

TypeScript is a tricky one. But I thought that including both might still be ok and will make it easy to use for both React as well as Angular as well as most other current framework's users (see [comment](https://github.com/andywer/webpack-blocks/issues/134#issuecomment-298836377))

sapegin:
I mean if you including TypeScript but excluding much more popular Sass it looks strange and inconsistent.

jvanbruegge:
Me personally would include Sass, but I'm not sure how many people use it

stevenjlho:
I think Sass is more popular than Typescript. For me, i want to add Sass package.

andywer:
Good points. Fortunately, we don't have to guess!

The sass-loader is in fact installed [~500k times per week](https://npm-stat.com/charts.html?package=sass-loader) (comparison: the super universal style-loader is installed [~800k times](https://npm-stat.com/charts.html?package=style-loader)).

So you are right, it is not just very popular, but extremely popular. We should add it, even it slows down the `npm install`.

andywer:
There we go. Added `sass` to `webpack-blocks` package as well 😉

andywer:
Do you want to have more time to review it or should I merge it and publish the alpha?

sapegin:
:shipit:

jvanbruegge:
for me it is ready to go

andywer:
Published under the `@next` tag. Check out [this comment in the 1.0 issue](https://github.com/andywer/webpack-blocks/issues/134#issuecomment-299906813) for further information.

Issue #152, opened 2017-05-08
resolveModules


diegohaz:
I'm currently using this snippet on my code: https://github.com/diegohaz/arc/blob/c066d6d0039fa5d782e8e3691f9d2bf3530a05bf/webpack.config.js#L36-L40

Since we have `resolveAliases` here, I wonder if it would be good to have a `resolveModules` block as well.

I can send a PR.

andywer:
Hey @diegohaz.

Yeah, might make sense to have a `resolveModules` in `@webpack-blocks/webpack`.

Or: Maybe just provide a `resolve()` to be used like `resolve({ modules: [ 'node_modules' ] })` or `resolve({ alias: { react: 'preact-compat' } })`?

@jvanbruegge

andywer:
@jvanbruegge What do you think?

jvanbruegge:
Sorry, somehow I missed the first notification.
I would say 👍 , and I would make `resolve` not `resolveModules`

BooleT37:
Some documentation would be nice

PR #160, opened 2017-05-19
Add resolve block (1.0)

diegohaz:
Related to #152 and #159

### DIFF
--- a/packages/webpack/README.md
+++ b/packages/webpack/README.md
@@ -88,11 +88,11 @@ Options object:
 }
 ```
 
-#### resolveAliases(aliases: object)
+#### resolve(config: object)
 
-Sets [resolve.alias](https://webpack.github.io/docs/configuration.html#resolve-alias). Use it to manually override module resolving.
+Sets [resolve](https://webpack.github.io/docs/configuration.html#resolve). Use it to manually override module resolving.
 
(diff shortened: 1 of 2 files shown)


andywer:
Thanks, @diegohaz :)

I just got one request: Removing `resolveAlias` from the README might be ok for 1.0, but it might make sense to leave the `resolveAlias` code in there for a little longer and just mark it as deprecated.

andywer:
Btw, are you currently using 0.4 or 1.0-alpha?

diegohaz:
I'm using 0.4

What's the approach you use to deprecate things? `console.warn`? Will do that later today.

andywer:
Add a `@deprecated` to the jsdoc block above the function and... a `console.warn()` makes sense :)

andywer:
But maybe we should list the possible properties `alias`, `modules`, ...

diegohaz:
Done.

Enhanced examples and updated webpack documentation url (we need to update it on other blocks).
https://webpack.js.org/configuration/resolve/

andywer:
Good job, @diegohaz. Thanks!

I think the `// import Utility from 'Utilities/utility'` comment is quite hard to understand, but that's just a minor thing. Maybe better `` // resolve `import 'Utilities/utility'` to correct path ``?

But that's really just a minor thing :)

PR #163, opened 2017-06-01
Major feature: match()

andywer:
This is quite a big pull request (sorry!), but it comes with a bunch of neat improvements, I think:

* Introduces `match()` to get rid of inconsistent `exclude`/`include` options in blocks
* API of most blocks didn't change or just dropped `exclude`/`include` options
* API of `@webpack-blocks/assets` (`css()`, `file()` & `url()` block) changed noticeably
* Docs have been updated

Needed to touch all blocks for this (😱), but I used this opportunity to improve the block tests a bit and add a few new tests. Also cleaned up some nasty code bits.

Please have a look, do a quick check of the changed code and read the updated docs if you can find some time. I would like to publish this as `1.0.0-beta`.

Shout if you disagree with something, I was out on a limb here 😉

### DIFF
--- a/packages/postcss/__tests__/integration.test.js
+++ b/packages/postcss/__tests__/integration.test.js
@@ -1,0 +1,93 @@
+import test from 'ava'
+import { css } from '@webpack-blocks/assets'
+import { createConfig, match } from '@webpack-blocks/core'
+import postcss from '../index'
+
+function LoaderOptionsPlugin (loaderOptions) {
+  this.loaderOptions = loaderOptions
+}
+
+test('Postcss works with defaults, without match()', t => {
+  const config = createConfig({}, [
+    postcss()
+  ])
+
+  t.deepEqual(config.plugins, [])
+  t.deepEqual(config.module.rules, [
+    {
+      test: /\.css$/,
+      use: [
+        'style-loader',
+        'css-loader',
+        {
+          loader: 'postcss-loader',
+          options: {}
+        }
+      ]
+    }
+  ])
+})
+
+test('Postcss works with css() & match()', t => {
+  const config = createConfig({}, [
+    match('*.css', { exclude: /node_modules/ }, [
+      css(),
+      postcss()
+    ])
+  ])
+
+  t.deepEqual(config.plugins, [])
+  t.deepEqual(config.module.rules, [
+    {
+      test: /^.*\.css$/,
+      exclude: /node_modules/,
+      use: [
+        'style-loader',
+        {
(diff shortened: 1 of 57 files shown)


andywer:
I know I touched a lot of stuff, but I hope you will find some time to quick-review the PR nevertheless.

If you are in a hurry, just check out the changed markdown files to see what changes for the user instead of reviewing the under-the-hood changes 😉

@jvanbruegge @sapegin @aaronjensen @eXon @diegohaz @bkonkle

packages/postcss/__tests__/integration.test.js
```diff
@@ -0,0 +1,93 @@
...
+        {
+          loader: 'postcss-loader',
+          options: {}
+        }
+      ]
+    }
+  ])
+})
+
+test('Postcss works with css() & match()', t => {
```
jvanbruegge: I would add a sass & match test case

andywer: Sure 👍

andywer: Done. But I added it to the global integration test (`packages/webpack/__tests__/integration.test.js`).

andywer:
Last chance to review this PR. Going to merge it today or tomorrow.

Will publish `v1.0.0-beta` thereafter 🎉

sapegin:
I think we should get rid of `context.fileType` entirely — I don’t see any benefits of having it now. We could just use globs everywhere and replace it with something like `context.testFor('*.css')`.

andywer:
Fair enough. I am not sure we even need a `context.testFor()`. Could just use `test: context.match.test || /\.css$/`.

In this case we would just change our loaders and deprecate `context.fileType`, yielding a warning if you still use it.

andywer:
What do you mean by "if it’s only for custom loaders"?

sapegin:
Blocks, not loaders ;-)

bkonkle:
>I think we should get rid of context.fileType entirely

I agree - there's not really a need for it any more, and it unnecessarily complicates things in my opinion. I like the `match()` api much better. 👍

andywer:
I pushed a commit to replace all usages of `context.fileType` in webpack-blocks and mention in the docs and code that it is deprecated.

I'd say this can now be published as 1.0 beta. We can introduce a warning message in the next version and remove it completely in the version thereafter.

bkonkle:
Awesome! 🎉

andywer:
Published as `v1.0.0-beta`

PR #165, opened 2017-06-06
v0.4: Update node-sass & sass-loader

### LINKED ISSUE #164: Remove `node-sass` from dependencies, update `sass-loader`
Looks like we need to bump `node-sass` to support Node 8: https://github.com/sass/node-sass/issues/1984

I tried to test out the change, but I don't see anywhere that `node-sass` is actually being used. Is it there just because of the peer dependency in `sass-loader`?

In any case, for Node 8 support we'll need to bump them to the latest versions. :-)

andywer:
Fixes #164.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,5 @@
 language: node_js
 node_js:
-  - 7
+  - 8
   - 6
   - 4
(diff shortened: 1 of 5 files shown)


andywer:
Published as `@webpack-blocks/sass v0.4.1`

Issue #142, opened 2017-04-12
Sass Modules


pakokrew:
Webpack blocks allow to use both global **SASS** and **CSS** modules, but we cannot use **SASS modules**.

We should either create a `sassModule` block, or add to `cssModule` the ability to load `x-sass` files and add `sass-loader`

I managed to use sass modules by merging css modules and sass configs :
```
{
    module:
      {
        loaders: [
          {
            test: context.fileType('text/x-sass'),
            loaders: [
              'style-loader',
              'css-loader?importLoaders=1&localIdentName=[name]--[local]--[hash:base64:5]&modules',
              'sass-loader',
            ]
          }
        ]
      }
  }
 ```

andywer:
Hey @pakokrew!

Good point, but not sure yet if it justifies a new block. Maybe it's fine just to add a second parameter to the `sass` block to pass `css-loader` options?

jvanbruegge:
This is basicly the same problem as #115 
Those blocks should both have a post hook doing that themself

andywer:
@jvanbruegge I don't think it's a similar issue, at least not the way it works right now. 115 is about moving it to a post hook, as you wrote, but here the post hook would not make a difference, since the `sass` block ships a separate `css-loader` instance for the *.sass/*.scss files we want to configure.

jvanbruegge:
But you pipe the result into a css loader: https://github.com/andywer/webpack-blocks/blob/master/packages/sass/index.js#L29
A post hook could add tge needed query string

andywer:
But why the post hook? We could leave it as it is and just add an options parameter for the css-loader option. Shouldn't make much difference, except that it's simpler. Or maybe I am missing something 😉

(Wouldn't add a post hook to the block unless it adds some real value)

jvanbruegge:
I mean a post hook on the css-modules block. So it can check all loaders for `css-loader` entries and add the query string there. So it _just works_ when using css-modules with sass or less

andywer:
Ahhh, now we are on the same page :)

Interesting thought, but we also have to consider that the user might not want to change *all* css-loaders. Even without the sass block you might want to load local css files as modules and css files from `node_modules/` as plain old css, for instance.

pakokrew:
I think we reach the point where the configuration is too specific to use cases, which maybe webpack blocks is not here to answer that but just helping "common" uses cases ...

andywer:
[...]

PR #168, opened 2017-06-22
Update SASS block readme

andywer:
Describe usage with CSS modules & PostCSS.


andywer:
@protoEvangelion Here is the PR for the updated docs. Did you have time to try it? :)

protoEvangelion:
@andywer Sure did but still trying to get it to work.  Probably making a silly error.

[Here is what I tried](https://github.com/andywer/webpack-blocks/issues/142#issuecomment-310850450).

Thanks!

PR #169, opened 2017-06-26
Make webpack a peer dependency, make webpack 3 compatible

### LINKED ISSUE #167: Migrate V1.0 to webpack 3
Webpack 3 was just released: https://github.com/webpack/webpack/releases/tag/v3.0.0
Only a few internal breaking changes so some plugins might be affected.

The big plus is scope hoisting like Rollup which increases initial paint and reduces bundle size

Would be nice to try

andywer:
Fixes #167

### DIFF
--- a/packages/webpack/package.json
+++ b/packages/webpack/package.json
@@ -19,8 +19,7 @@
   "bugs": "https://github.com/andywer/webpack-blocks/issues",
   "dependencies": {
     "@webpack-blocks/core": "^1.0.0-beta",
-    "assert-plus": "^1.0.0",
-    "webpack": "^2.6.1"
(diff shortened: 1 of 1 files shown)


Vlad-Zhukov:
Hello, what do you think about also removing webpack from the exports?

https://github.com/andywer/webpack-blocks/blob/release/1.0/packages/webpack/index.js#L18

andywer:
Could probably be removed. Are there any strong feelings to remove it?

Vlad-Zhukov:
Just a common sense.

andywer:
I agree, but will fix it in a separate PR.

andywer:
See #170

PR #177, opened 2017-07-19
Fix resolve block to prefer custom extensions over defaults

### LINKED ISSUE #176: resolve.extensions cannot be configured properly
From @TombolaShepless in #153:

> We have `resolve.extensions` setup so that we can generate different bundles based on device types (e.g. mobile, tablet etc.). However because these two values appear first, Webpack will always resolve the "base" file rather than the override:
>
> `import foo from './foo'` <--- If we have a `foo.mobile` file it will never be loaded.
>
> Would it be safer to make no assumptions and have this as an empty array? Or is there an option/method I am missing? Prior to v1 we were using the create vanilla config approach but it looks like this has been removed?

zcei:
Allow custom extensions configurations, like
```js
resolve({
  extensions: ['.custom.js']
})
```
so that 
```js 
import foo from './foo'
```
first checks `./foo.custom.js` before falling back to `./foo.js`

Closes #176

### DIFF
--- a/packages/webpack/__tests__/integration.test.js
+++ b/packages/webpack/__tests__/integration.test.js
@@ -1,6 +1,6 @@
 import test from 'ava'
 import path from 'path'
-import { createConfig, entryPoint, match, setOutput, sourceMaps } from '../index'
+import { createConfig, entryPoint, match, setOutput, sourceMaps, resolve } from '../index'
 import { css, file, url } from '@webpack-blocks/assets'
 import babel from '@webpack-blocks/babel6'
 import devServer from '@webpack-blocks/dev-server'
@@ -176,4 +176,16 @@ test('context contains necessary properties', t => {
       return prevConfig => prevConfig
     }
   ])
+})
+
+test('prepends custom extension to default ones', t => {
+  const expectedExtensionOrder = ['.custom.js', '.js', '.json']
+
+  const webpackConfig = createConfig([
+    resolve({ extensions: ['.custom.js'] })
+  ])
+
+  const actualExtensions = webpackConfig.resolve.extensions
+
+  t.deepEqual(actualExtensions, expectedExtensionOrder)
 })
--- a/packages/webpack/index.js
+++ b/packages/webpack/index.js
@@ -7,6 +7,7 @@
 const assert = require('assert-plus')
 const core = require('@webpack-blocks/core')
 const webpack = require('webpack')
+const webpackMerge = require('webpack-merge')
 const path = require('path')
 const parseVersion = require('./lib/parseVersion')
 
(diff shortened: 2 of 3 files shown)


packages/core/lib/blockUtils.js
```diff
@@ -1,5 +1,6 @@
 const _ = require('lodash')
 const webpackMerge = require('webpack-merge')
+const webpackBlocksMerge = webpackMerge.smartStrategy({ 'resolve.extensions': 'prepend' })
```
zcei: For now I left it inlined, but in case there will be more custom behavior I'd recommend moving this object to something like `merge-strategies.js`

andywer: I would say we should rather put this into [resolve()](https://github.com/andywer/webpack-blocks/blob/release/1.0/packages/webpack/index.js).

andywer:
Thanks for your PR, @zcei!

packages/core/lib/blockUtils.js
zcei: When putting it to `resolve`, we start introducing a lot of small different merge behaviors.
`webpack-blocks` allows you to define your own strategies and I think it makes sense to make use of it, as it alters only the object paths defined.

But yeah, we can still refactor this in case there are a lot of different merging strategies we want to have. 
You could also argue as well that this isn't core behavior and some people might want to write something that actually replaces / appends / do other crazy stuff.

Going to change ✅

zcei:
@andywer I exposed a new util fn, that allows merging with given strategy, to not duplicate the `webpack-merge` dependency from `core` into the `webpack` block.

And it makes it very clear in every single block how you alter the default behavior.

WDYT?

andywer:
I get your intention, but my personal aim is to keep the utils as slim as possible. As you might know there is already a long-running discussion if util functions should be passed to the blocks at all.

Duplicating the webpack-merge dependency to `@webpack-blocks/webpack` might not look to elegant, but still seems like the smaller evil, compared to extending the utils by a function that is only used in one spot.

After all: No one said you have to stick to the utils to do the merging. In fact the 1.0 API blackboxes the way a block applies its changes to the config, so if you need something special (like here), feel free to do whatever you need to do in the block :)

andywer:
Btw, sorry for the wall of text 😉

I would open a PR with my solution and let you review it as soon as I find a few minutes time, but I wanted to wait for your response/feedback or that of my fellow co-contributors first.

zcei:
Yes, I was actually looking into putting it into the `webpack` block, but wanted to write a test for it then there as well and didn't get enough time to do so.

Happy to review other approaches, but better not double the work 🙂

andywer:
Sure, let's not double the work. Update the PR when you got time. But if you are just lacking the unit tests, don't mind; I would add them as well :)

zcei:
Sorry, was a bit short on time, but pushed what I got now, including a bare-minimal test to actually check the prepend behavior.

PR #178, opened 2017-07-24
Better error handling of configSetters param

sapegin:
Validate `configSetters` in all public API functions.


andywer:
👍

Fyi: Last PR before I will be offline for the next few days 😉

PR #183, opened 2017-08-04
Add package-lock.json files, update dependencies

andywer:
`package-lock.json` is a new feature of npm 5 and works roughly like `yarn.lock`.


sapegin:
Just like Yarn but better ;-)

PR #186, opened 2017-08-08
Add UglifyJS block

sapegin:
Should I add it to `webpack-blocks` package too?

### DIFF
--- a/packages/uglify/README.md
+++ b/packages/uglify/README.md
@@ -11,11 +11,12 @@ Based on [uglifyjs-webpack-plugin](https://github.com/webpack-contrib/uglifyjs-w
 ## Usage
 
 ```js
-const { createConfig } = require('@webpack-blocks/webpack')
-const uglify = require('@webpack-blocks/uglify')
+const { createConfig, env, uglify } = require('webpack-blocks')
 
 module.exports = createConfig([
-  uglify(/* options */)
+  env('production', [
+    uglify(/* options */)
+  ])
 ])
 ```
 
@@ -35,8 +36,6 @@ By default generates this configuration:
         cache: true
       },
       uglifyOptions: {
-        ie8: false,
-        ecma: 5,
         compress: {
           warnings: false
         }
--- a/packages/uglify/README.md
+++ b/packages/uglify/README.md
@@ -1,0 +1,55 @@
+# Webpack blocks - UglifyJS
+
+[![JavaScript Style Guide](https://img.shields.io/badge/code%20style-standard-brightgreen.svg)](http://standardjs.com/)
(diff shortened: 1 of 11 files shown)


jvanbruegge:
I think this block is not really needed, as it's basicly just
```js
addPlugins([
    new webpack.optimize.uglifyjs(options)
]);
```

sapegin:
In this case you’ll have to copy paste options to each project, and defaults are far from ideal. Also it’s not `webpack.optimize.UglifyJsPlugin` but uglifyjs-webpack-plugin that supports ES6.

packages/uglify/README.md
```diff
@@ -0,0 +1,55 @@
...
+```js
+{
+  plugins: [
+    new UglifyJSPlugin({
+      parallel: {
+        cache: true
+      },
+      uglifyOptions: {
+        ie8: false,
+        ecma: 5,
```
jvanbruegge: If you want to support ES6, what does this do?

jvanbruegge:
I see

packages/uglify/package.json
```diff
@@ -0,0 +1,33 @@
+{
+  "name": "@webpack-blocks/uglify",
+  "version": "1.0.0",
+  "description": "Webpack block for UglifyJS",
+  "main": "index",
+  "license": "MIT",
+  "author": "Andy Wermke <andy@dev.next-step-software.com>",
```
andywer: Feel free to put your name in here 😉

andywer:
Yes, that should then be part of the `webpack-blocks` package as well.

packages/uglify/package.json
sapegin: :shipit:

packages/uglify/README.md
sapegin: I read the docs again: we should leave the default values here: 8 for the parser and 5 for the printer.

sapegin: And ie8 is false by default too ;-|

sapegin:
@andywer All done! 🍕

andywer:
Cool!

@sapegin No intention to add it to the `webpack-blocks` package immediately? 😉

sapegin:
@andywer Done ;-)

packages/uglify/README.md
```diff
@@ -0,0 +1,54 @@
...
+Based on [uglifyjs-webpack-plugin](https://github.com/webpack-contrib/uglifyjs-webpack-plugin) (not `webpack.optimize.UglifyJsPlugin`) which uses UglifyJS v3 (uglify-es) that supports ECMAScript 2015.
+
+
+## Usage
+
+```js
+const { createConfig, env, uglify } = require('webpack-blocks')
+
+module.exports = createConfig([
+  env('production', [
```
andywer: 👍 for using `env('production', [])` in the sample

andywer:
@andywer Memo to myself: Don't forget to publish.

PR #187, opened 2017-08-08
Less noisy base config

sapegin:
Do not print long list of chunks etc. on build.

### DIFF
--- a/packages/dev-server/index.js
+++ b/packages/dev-server/index.js
@@ -48,9 +48,8 @@ function postConfig (context, util) {
         hotOnly: true,
         historyApiFallback: true,
         inline: true,
-        // Disable verbose loggin in browser’s console
-        // It will still show compile warnings and errors
-        clientLogLevel: 'none',
+        // Disable verbose logging in browser’s console, only print errors
+        clientLogLevel: 'error',
         // Do not print chunks list on every compilation, only print errors
         stats: 'errors-only'
       }, context.devServer.options),
--- a/packages/dev-server/index.js
+++ b/packages/dev-server/index.js
@@ -49,6 +49,7 @@ function postConfig (context, util) {
         historyApiFallback: true,
         inline: true,
         // Disable verbose loggin in browser’s console
+        // It will still show compile warnings and errors
         clientLogLevel: 'none',
         // Do not print chunks list on every compilation, only print errors
         stats: 'errors-only'
--- a/packages/dev-server/index.js
+++ b/packages/dev-server/index.js
@@ -47,7 +47,11 @@ function postConfig (context, util) {
         hot: true,
         hotOnly: true,
         historyApiFallback: true,
-        inline: true
+        inline: true,
+        // Disable verbose loggin in browser’s console
+        clientLogLevel: 'none',
+        // Do not print chunks list on every compilation, only print errors
+        stats: 'errors-only'
       }, context.devServer.options),
       entry: addDevEntryToAll(prevConfig.entry || {}, entryPointsToAdd),
       plugins: [
--- a/packages/core/CHANGELOG.md
+++ b/packages/core/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## Next version
 
 - More useful error message when passing invalid blocks to `createConfig()` ([#171](https://github.com/andywer/webpack-blocks/issues/171))
+- Less noisy default options
 
 ## 1.0.0-beta
 
--- a/packages/core/lib/__tests__/createConfig.test.js
+++ b/packages/core/lib/__tests__/createConfig.test.js
@@ -3,6 +3,26 @@ import sinon from 'sinon'
 import { createConfig } from '../index'
 import blockHelpers from '../blockUtils'
 
+const defaultConfig = {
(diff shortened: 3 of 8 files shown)


andywer:
The [webpack documentation](https://webpack.js.org/configuration/stats/) says "for webpack-dev-server, this property needs to be in the `devServer` object".

So I suppose we have to add this to the dev-server block code as well?

sapegin:
Yup, that would be my next PR ;-)

andywer:
But wouldn't it make sense to do both fixes in one PR (one "less noise" PR)?

sapegin:
I also wanted to enable error overlay and compression for devServer. Can include that as well, or send as another PR.

sapegin:
And this as well: https://github.com/facebookincubator/create-react-app/blob/2069e8ac43c4765f1698091d3985ecb70ba88089/packages/react-scripts/config/webpackDevServer.config.js#L75-L79

andywer:
Proposal: I would rather not put `node_modules` exclusion and compression in the dev-server as an implicit default. The error overlay (you mean the red error screen?) maybe, maybe not. But since you don't expect webpack-blocks to silently change your config I would not do so, unless it's extremely generic and extra-unlikely to ever cause any issues.

andywer:
But +1 for the proposed changes. Just not a big fan of letting very generic blocks change detail configuration, since that already caused multiple issues in the project's past.

sapegin:
I think you see webpack-blocks more low lever than it should be to be really useful ;-)

> I would rather not put node_modules exclusion

Why not? People don’t change code inside node_modules often.

> and compression in the dev-server as an implicit default.

What downsides you see here? All modern browsers support compression and it would make it a little bit faster. I think it would be a good default for most users.

> The error overlay (you mean the red error screen?) maybe, maybe not.

Yes, it makes DX much better — you don’t see an app that looks normal but wasn’t even compiled ;-)

> But since you don't expect webpack-blocks to silently change your config I would not do so, unless it's extremely generic and extra-unlikely to ever cause any issues.

I think this is the place where our visions differ the most ;-) That’s exactly what I want webpack-blocks to do and why I use it! And there’s no my config, only config that was generated by webpack-blocks. I think we shouldn’t expect webpack-blocks to know all webpack defaults but should try to find our own default that would make the best experience for most of them. Similar to Create React App but we’re in a better position here: our users could easily change defaults ;-)

andywer:
Maybe it is 😉

`node_modules/` exclusion:
Just cautious, since it might cause issues for people who are used to do quick hacks in their node_modules and now wonder why it won't update anymore. Otherwise fine to me.

Compression:
Yeah, compression will probably be fine, ignore my comment :)

Error screen:
There is just one thing about it that might annoy people (thus would make it a rather personal thing). When developing with HMR if you change the code and it breaks it will replace your currently running site with the error screen. Without it you are able to use the site normally, just see the error in the console and everything will resume normally as soon as you fix the error.

Don't get me wrong, I like your practical view on usage! It's just that I became quite cautious about defaults in generic blocks. Providing higher level abstractions on top of low-level stuff is easy, changing low-level things that annoy you when all you have got is high-level blocks, is not.

sapegin:
>  it might cause issues for people who are used to

This could be said about any non-default value we have ;-) I think we should document default config for each block and it would solve this and some other issues (I have to read blocks’ sources almost every time now). But anyway with normal webpack I had to read the docs every time because I think people are used to how *their* app behave and not to defaults.

> Without it you are able to use the site normally

This is probably too opinionated to decide between two of us — we need more people here. On your side is the webpack default value, on my — CRA ;-)

sapegin:
@andywer I think it’s ready to be merged ;-) Everything else is out of the scope of this PR anyway.

andywer:
How about asking the twitter followers what they'd prefer? The users don't necessarily know what they want, but should be ok to make an educated decision here.

sapegin:
@andywer Do it and I’ll retweet ;-)

packages/dev-server/index.js
```diff
@@ -47,7 +47,11 @@ function postConfig (context, util) {
         hot: true,
         hotOnly: true,
         historyApiFallback: true,
-        inline: true
+        inline: true,
+        // Disable verbose loggin in browser’s console
+        clientLogLevel: 'none',
```
andywer: I'd rather prefer `warn` or at least `error` here. I don't always look at the console running webpack. The on-page console is open pretty much all-the-time instead.

sapegin: > Note that the console will always show bundle errors and warnings. This option only effects the message before it.

https://webpack.js.org/configuration/dev-server/#devserver-clientloglevel

andywer: What do they mean by "the message before it"? 🤔

sapegin: I’ve experimented a bit. This paragraph [is misleading](https://github.com/webpack/webpack-dev-server/issues/1038). Changing it to `error`.

andywer: 👍👍

andywer:
Just one more thing (sorry!): Could you please update the changelogs?

You can just use "Next version" as version title.

PR #188, opened 2017-08-09
[WIP] New vendor bundle block

sapegin:


diegohaz:
Related: https://github.com/diegohaz/webpack-blocks-split-vendor

sapegin:
@diegohaz I think both are useful because there are different: yours is more generic and mine is more automated. 

@andywer What do you think? ;-)

diegohaz:
I think both are (or could be) the same. I'm all for having an official block and deprecating mine.

I would just check if the problem I wrote there still persists: https://github.com/diegohaz/webpack-blocks-split-vendor#how-it-does (the reason I used `webpack-md5-hash`).

packages/vendor-bundle/README.md
```diff
@@ -0,0 +1,60 @@
...
+
+Name of the main bundle of your app.
+
+#### filename *(default: vendor.[chunkhash].js)*
+
+Output bundle file name.
+
+#### minChunks *(default: 5)*
+
+Minimal number of chunks to include dependency into the bundle.
```
andywer: Why wouldn't you want to include a dependency with few chunks into the vendor bundle?

andywer:
I think the md5-hashing is just an alternative solution to what @sapegin implemented as `runtimeBundle`. Not sure what's better.

Thanks for pointing out your package, @diegohaz. I almost forgot about it.

I think the way this package (in this PR) works seems cleaner, but from a high-level view they do pretty much exactly the same, right?

diegohaz:
What's `runtimeBundle`?

sapegin:
> I think the md5-hashing is just an alternative solution to what @sapegin implemented as runtimeBundle. Not sure what's better.

Yup, it’s solving the same problem. As I understand the benefit of md5-hashing is that you don’t have another bundle just for IDs. 

> What's runtimeBundle?

Here it is:

```js
/**
 * Extract webpack runtime with module IDs so changes in app code won’t invalidate the vendor bundle.
 *
 * @returns {function}
 */
module.exports = function() {
    return (context, { addPlugin }) => addPlugin(
        new context.webpack.optimize.CommonsChunkPlugin({
            name: 'runtime',
            minChunks: Infinity,
        })
    );
};
```

diegohaz:
I see now (didn't find it in the code though). That's pretty neat.

packages/vendor-bundle/README.md
sapegin: Because you may have some big dependencies that you’re using only on several pages, like some fancy chart library.

sapegin: So my idea was to include dependencies that are used on most pages.

andywer: Ahhh, it's not about the size of the dependency, but how many chunks are using this dependency. I think the documentation should highlight that 😉

sapegin: But that’s exactly what is says! ;-)

andywer: At least I didn't get that it means "number of chunks using the dependency" and I am maintainer 😜

sapegin: OK, I’ll update it ;-)

andywer:
We still need a decision on that, right?

sapegin:
I want to make it more generic and support both use cases: my initial and from @diegohaz’s block so we could have one block “to rule them all” ;-)

sapegin:
Looks like this needs more research. Reading:

https://medium.com/webpack/predictable-long-term-caching-with-webpack-d3eee1d3fa31
https://github.com/facebookincubator/create-react-app/issues/2328
https://github.com/facebookincubator/create-react-app/pull/2310

marcofugaro:
Whoa, good idea with this PR, vendor splitting is the most common and most reasonable code splitting.

I did this once to a big project to minimize dev build times, I also tried to implement the [DllPlugin](https://webpack.js.org/plugins/dll-plugin/) on the vendor chunk to basically freeze it and almost remove the compile times of that chunk, but I wasn't able to do it because of the too much overhead of that plugin, but maybe you guys can do it!

sapegin:
An one more link:

https://iamakulov.com/notes/webpack-front-end-size-caching/

vlad-zhukov:
I'd like all options to be passed down to a [CommonsChunkPlugin](https://webpack.js.org/plugins/commons-chunk-plugin/) rather than a small whitelist.

sapegin:
@vlad-zhukov Yeah, I’m going to make it more generic.

packages/vendor-bundle/index.js
```diff
@@ -0,0 +1,54 @@
...
+    filename: 'vendor.[chunkhash].js',
+    minChunks: 1
+  }, options)
+
+  return (context, util) => {
+    const plugin = new webpack.optimize.CommonsChunkPlugin({
+      name: options.main,
+      filename: options.filename,
+      children: true,
+      minChunks ({ userRequest }, count) {
```
dmitmel: `resource` is better than `userRequest` because it's more accurate.

packages/vendor-bundle/index.js
```diff
@@ -0,0 +1,54 @@
...
+        return userRequest && userRequest.includes('node_modules') &&
```
dmitmel: `String#indexOf` is faster than `String#includes` (see this [benchmark](http://jsbench.github.io/#a4612afd0cd26e911ee8).

packages/vendor-bundle/package.json
```diff
@@ -0,0 +1,31 @@
...
+    "chunk"
+  ],
+  "repository": "andywer/webpack-blocks",
+  "bugs": "https://github.com/andywer/webpack-blocks/issues",
+  "dependencies": {},
+  "devDependencies": {
+    "@webpack-blocks/core": "^1.0.0-beta",
+    "ava": "^0.18.2",
+    "standard": "^8.6.0",
+    "webpack": "^3.5.1"
```
dmitmel: `webpack` must be in `peerDependencies` because `index.js` imports it.

marcofugaro:
Just a reminder, [webpack 4 splits vendors by defult](https://gist.github.com/sokra/1522d586b8e5c0f5072d7565c2bee693), so this functionality will be already there when webpack-blocks is upgraded wo webpack 4

sapegin:
This block doesn't make any sense in webpack 4, I'm going to close the PR.

PR #192, opened 2017-08-12
Add a list of third-party blocks

sapegin:
- List of third-party blocks
- Fix title casing
- Update TOC
- Add missed Elm block

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,17 +12,17 @@ Ready to use blocks to configure popular tools like *Babel*, *PostCSS*, *Sass*, 
 >[Dan Abramov](https://github.com/gaearon) via [twitter](https://twitter.com/dan_abramov/status/806249934399881216) (Co-author of Redux, Create React App and React Hot Loader)
 
 
-## Table of Contents
+## Table of contents
 
 - [Installation](#installation)
 - [Example](#example)
-- [More Examples](#more-example)
-- [Custom Blocks](#custom-blocks)
-- [Available Webpack Blocks](#available-webpack-blocks)
-- [Third-party Blocks](#third-party-blocks)
-- [Design Principles](#design-principles)
+- [More examples](#more-example)
+- [Custom blocks](#custom-blocks)
+- [Available webpack blocks](#available-webpack-blocks)
+- [Third-party blocks](#third-party-blocks)
+- [Design principles](#design-principles)
 - [FAQ](#faq)
-- [Like What You See?](#like-what-you-see)
+- [Like what you see?](#like-what-you-see)
 - [License](#license)
 
 
@@ -98,7 +98,7 @@ module.exports = createConfig([
 
 All blocks, like `babel` or `postcss` are also available as their own [small packages](./packages), webpack-blocks package wraps them as a single dependency for convenience.
 
-## More Examples
+## More examples
 
 CSS modules:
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -15,11 +15,14 @@ Ready to use blocks to configure popular tools like *Babel*, *PostCSS*, *Sass*,
...
-- [Design principles](#design-principles)
+- [Example](#example)
+- [More Examples](#more-example)
+- [Custom Blocks](#custom-blocks)
+- [Available Webpack Blocks](#available-webpack-blocks)
+- [Third-party Blocks](#third-party-blocks)
+- [Design Principles](#design-principles)
 - [FAQ](#faq)
-- [Like what you see?](#like-what-you-see)
+- [Like What You See?](#like-what-you-see)
```
jvanbruegge: I think that's weird, as it is a normal sentence

sapegin: No, it’s a heading, so it should be capitalized as all other headings. Personally I never use title case but I see that @andywer is using it in other documents.

andywer: Mixed feeling. Looks a bit strange indeed, but inconsistency is annoying as well... Not sure, mixed feelings about both our options 😅

PS: Same applies for "Available Webpack Blocks" (strange to see uppercase webpack, esp. if it's not even the first word in a sentence.

sapegin:
The best solution in my opinion is not to use title casing at all ;-) It’s harder to read, it doesn't look good and it’s [hard to do right](http://titlecapitalization.com/).

For example [Google suggests](https://material.io/guidelines/style/writing.html) to use sentence case everywhere in the UI.

andywer:
We can also stick to sentence-style capitalization, I don't mind.

@jvanbruegge @zcei?

jvanbruegge:
I'm in favor of sentence captitalization. Spelling everything uppercase just looks weird

sapegin:
OK, three 👍  for ~~sane~~ sentence casing, I’ll update the PR and send a new one to change the reset of the docs.

sapegin:
Done.

README.md
```diff
@@ -163,16 +166,28 @@ Check out the [sample app](./test-app) to see a webpack config in action or read
...
+- [typescript](./packages/typescript)
 - [uglify](./packages/uglify)
 - [webpack](./packages/webpack)
 
+
+## Third-party blocks
+
+- [webpack-blocks-happypack](https://github.com/diegohaz/webpack-blocks-happypack) — HappyPack
+- [webpack-blocks-purescript](https://github.com/ecliptic/webpack-blocks-purescript) — PureScript
+- [webpack-blocks-server-source-map](https://github.com/diegohaz/webpack-blocks-server-source-map) — source map for server bundle
```
andywer: One idea before I approve: Might make sense to upper-case the first word of the description for a more consistent look ("- Source map for server bundle", "- Vendor bundle")

sapegin: Hm, but it’s not a new sentence…

andywer: Just thought that it might look more consistent, since `HappyPack`, `PureScript`, `TypeScript`, ... all start with upper-case letters.

sapegin: Because they are names ;-)

andywer: Sure, I just mean that it looks a tiny bit odd. But I don't really mind, just wanted to get a 2nd opinion 😉

sapegin: That’s because you used to semirandomly capitalized words in German ;-)

PR #193, opened 2017-08-17
Remove title casing everywhere

sapegin:


andywer:
Thanks a bunch, @sapegin, for eagerly fixing that annoying sh** most people don't even touch! 😅

I owe you 😉

PR #197, opened 2017-08-24
v1.0.0-beta.4

andywer:


packages/elm/CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # @webpack-blocks/elm - Changelog
 
+## 1.0.0-beta.2
+
+- Added `package-lock.json`
```
sapegin: Why is this in the change log? I mean why users should care about changes they will never see? ;-)

andywer: Just wanted to provide a reason why there is a new version 😉

sapegin: Right answer: there shouldn’t be a new version because there are no changes that may affect users ;-)

andywer: I was aiming for consistency. The `package-lock.json` makes sense and I would like to publish it for all core packages at once.

sapegin: It make sense for us but not for users, this shouldn’t be in the change log. It won’t be even published to npm.

andywer: So what should we do to publish a new version, but without this line in the changelog? Should I put it in parenthesis? ^^

sapegin: But what’s the reason to publish a new version without any changes?

andywer: Ahh, my bad. Sorry, I got something wrong about the `package-lock.json`.
For future reference: Here is its [documentation](https://docs.npmjs.com/files/package-lock.json).

In this case it doesn't make sense to bump the version, you are right 👍

sapegin: 🍕

sapegin:
Are you going to finish and merge this? ;-)

andywer:
I reverted the "package-lock.json only" version bumps and the build is now green again after I restarted it last night (wtf?).

So it can be merged and published as soon as the tests for the latest commit have run.

sapegin:
It was failing because of Elm packager issues — I’ve restarted it in the morning it it’s fine now (in my PR as well).

andywer:
Published!

PR #202, opened 2017-08-29
Add a note about enabling hot reload

sapegin:


andywer:
Good point, we need to address this. Still not sure, though, if we should use `react-hot-loader/webpack` or not.

sapegin:
This is not about react-hot-loader just a basic webpack hot reload.

I think we should deprecate react-hot-loader and suggest users to create a third-party block.

andywer:
Yeah, I agree on that. Was just referring to that because you chose a react hot reloading example 😉

sapegin:
Yeah, because that’s a sample from my article and I’ve spent some time figuring out why it’s not working ;-)

sapegin:
Do you want me to change anything here?

andywer:
No, it's fine. Thx for reminding me :)

PR #203, opened 2017-08-30
Mention error overlay in the example

sapegin:


packages/dev-server/README.md
```diff
@@ -16,7 +16,10 @@ const devServer = require('@webpack-blocks/dev-server')
 module.exports = createConfig([
   // use only if `NODE_ENV === 'development'`:
   env('development', [
-    devServer(),
+    devServer({
+        // Show error overlay in the browser
```
andywer: Ahh, really. Had to look-up that this is a built-in functionality now. Maybe "Show error screen in the browser"? "Overlay" always sounds so cryptic. Or is it just me? ^^

sapegin: I think it’s a common name for the thing now. “Screen” sounds too generic for me ;-)

andywer: How about "Show error in the application"?

sapegin: I still want to keep “error overlay” because that’s the name of the think and developers know it as “error overlay”. Otherwise I don’t mind ;-)

sapegin: Let’s copy from webpack docs then:

> Shows a full-screen overlay in the browser when there are compiler errors or warnings. 

😝

andywer: Sure 😄. Might wanna shorten it a bit: "Show full-screen overlay in the browser on compiler errors or warnings"?

sapegin: Done ;-)

PR #206, opened 2017-09-06
Add setEnv block

### LINKED ISSUE #201: setEnv block
I assume that most of the `defineConstants` block usages are like this:

```js
defineConstants({
  'process.env.NODE_ENV': process.env.NODE_ENV
}),
```

So it would make sense to have a wrapper around [webpack.EnvironmentPlugin](https://webpack.js.org/plugins/environment-plugin/):

```js
setEnv(['NODE_ENV', 'DEBUG'])

setEnv({
  NODE_ENV: 'development', // use 'development' unless process.env.NODE_ENV is defined
  DEBUG: false
})
```

I know @andywer isn’t a fan of this idea but I’m sure I could find some supporters 👻 

---

The code would be similar to [defineConstants](https://github.com/andywer/webpack-blocks/blob/release/1.0/packages/webpack/lib/defineConstants.js).

Example implementation: https://github.com/vlad-zhukov/webpack-blocks-more/blob/master/src/setEnv.js

Please send your pull request to `release/1.0` branch.

vlad-zhukov:
Added a `setEnv` block as per discussion. Resolves #201

### DIFF
--- a/packages/webpack/README.md
+++ b/packages/webpack/README.md
@@ -106,7 +106,23 @@ Replaces constants in your source code with a value (`process.env.NODE_ENV` for 
 
 Every constant's value is `JSON.stringify()`-ed first, so you don't have to remember.
 
-Special feature: Using `defineConstants` multiple times results in a single DefinePlugin instance configured to do all the replacements.
+Using `defineConstants` multiple times results in a single DefinePlugin instance configured to do all the replacements.
+
+#### setEnv(constants: string[]|object): Function
+
+Replaces constants in your source code with a values from `process.env` using the [webpack.EnvironmentPlugin](https://webpack.js.org/plugins/environment-plugin/).
+
+Using `setEnv` multiple times results in a single EnvironmentPlugin instance configured to do all the replacements.
+
+```js
+module.exports = createConfig([
+  setEnv(['NODE_ENV']),
+  setEnv({
+    BABEL_ENV: 'development', // use 'development' unless process.env.BABEL_ENV is defined
+    PORT: 3000,
+  }),
+])
+```
 
 #### entryPoint(entryPoint: string|string[]|object)
 
--- a/packages/webpack/lib/setEnv.js
+++ b/packages/webpack/lib/setEnv.js
@@ -1,0 +1,40 @@
+module.exports = setEnv
+
+/**
+ * Replaces constants in your source code with values from `process.env`
+ * using the `webpack.EnvironmentPlugin`.
+ *
+ * Using `setEnv` multiple times results in a single
+ * EnvironmentPlugin instance configured to do all the replacements.
+ *
+ * @param {string[]|object} constants
+ * @return {Function}
+ */
+function setEnv (constants) {
+  const setter = context => prevConfig => {
+    context.setEnv = Object.assign({}, context.setEnv, toObject(constants))
+    return prevConfig
+  }
+
+  return Object.assign(setter, { post: addEnvironmentPlugin })
+}
+
+function addEnvironmentPlugin (context, util) {
+  return util.addPlugin(
+    new context.webpack.EnvironmentPlugin(context.setEnv)
+  )
+}
+
+function toObject (constants) {
+  if (Array.isArray(constants)) {
+    return constants.reduce((result, constant) => {
+      // Setting the default (fallback) value to `undefined`
+      // to make sure the EnvironmentPlugin will throw a warning
+      // in case if `process.env[constant]` was not defined too.
+      result[constant] = undefined
+      return result
+    }, {})
+  }
+
+  return constants
+}
--- a/packages/webpack/CHANGELOG.md
+++ b/packages/webpack/CHANGELOG.md
@@ -1,5 +1,9 @@
 # @webpack-blocks/webpack - Changelog
 
+## Next version
+
+- Added `setEnv()` ([#206](https://github.com/andywer/webpack-blocks/pull/206))
+
 ## 1.0.0-beta.3
 
 - Make resolve() prepend custom extensions ([#177](https://github.com/andywer/webpack-blocks/issues/177))
--- a/packages/webpack/index.js
+++ b/packages/webpack/index.js
@@ -21,6 +21,7 @@ exports.createConfig = createConfig
 exports.addPlugins = addPlugins
 exports.customConfig = customConfig
 exports.defineConstants = require('./lib/defineConstants')
+exports.setEnv = require('./lib/setEnv')
 exports.entryPoint = entryPoint
 exports.performance = performance
 exports.resolve = resolve
--- a/packages/webpack/lib/__tests__/setEnv.test.js
+++ b/packages/webpack/lib/__tests__/setEnv.test.js
@@ -1,0 +1,19 @@
+const test = require('ava')
+const { createConfig } = require('@webpack-blocks/core')
+const setEnv = require('../setEnv')
+
+const webpack = {
+  EnvironmentPlugin: function EnvironmentPluginMock (definitions) {
+    this.definitions = definitions
+  }
+}
+
+test('setEnv() creates a single EnvironmentPlugin instance only', (t) => {
+  const config = createConfig({ webpack }, [
+    setEnv(['a']),
+    setEnv({b: 'foo'})
+  ])
+
+  t.is(Object.keys(config.plugins).length, 1)
+  t.deepEqual(config.plugins[0].definitions, {a: undefined, b: 'foo'})
+})
--- a/packages/webpack/lib/defineConstants.js
+++ b/packages/webpack/lib/defineConstants.js
@@ -5,7 +5,7 @@ module.exports = defineConstants
  * for example) using the `webpack.DefinePlugin`. Every constant's value is
  * `JSON.stringify()`-ed first, so you don't have to remember.
  *
- * Special feature: Using `defineConstants` multiple times results in a single
+ * Using `defineConstants` multiple times results in a single
  * DefinePlugin instance configured to do all the replacements.
  *
  * @param {object} constants  { [constantName: string]: * }


packages/webpack/README.md
```diff
@@ -81,6 +81,12 @@ Every constant's value is `JSON.stringify()`-ed first, so you don't have to reme
 
 Special feature: Using `defineConstants` multiple times results in a single DefinePlugin instance configured to do all the replacements.
 
+#### setEnv(constants: string[]|object): Function
+
+Replaces constants in your source code with a values from `process.env` using the [webpack.EnvironmentPlugin](https://webpack.js.org/plugins/environment-plugin/).
+
+Special feature: Using `setEnv` multiple times results in a single EnvironmentPlugin instance configured to do all the replacements.
```
sapegin: - “Special feature” → “Note”.
- Couple of examples would be nice to have here.

packages/webpack/lib/setEnv.js
```diff
@@ -0,0 +1,26 @@
...
+ *
+ * Special feature: Using `setEnv` multiple times results in a single
+ * EnvironmentPlugin instance configured to do all the replacements.
+ *
+ * @param {string[]|object} constants
+ * @return {Function}
+ */
+function setEnv (constants) {
+  const setter = context => prevConfig => {
+    context.setEnv = Object.assign({}, context.setEnv, constants)
```
sapegin: Would it work this way?

```js
createConfig([
  setEnv({ NODE_ENV: 'development' }),
  setEnv(['DEBUG'])
])
```

sapegin:
And thanks for the pull request! ;-)

packages/webpack/lib/setEnv.js
vlad-zhukov: It wouldn't I guess, do we need a multi-setEnv functionality or it can be simplified to a straight `EnvironmentPlugin` call?

packages/webpack/README.md
vlad-zhukov: Copied directly from the `defineConstants` block above.

packages/webpack/lib/setEnv.js
sapegin: Not sure, in this case we’ll need to parse options ourselves.

@andywer What do you think?

packages/webpack/README.md
sapegin: Then we’ll need to change it there too, there’s nothing *special* that, just a little convenience ;-)

packages/webpack/lib/setEnv.js
andywer: We should definitely support having multiple `setEnv()`. And having the ability to pass an array like `setEnv(['DEBUG'])` that is resolved to `setEnv({ DEBUG: process.env.DEBUG })` would also be great.

I hope I got your question right.

andywer: And thanks for the PR, @vlad-zhukov! :)

sapegin: The question is should we try to merge arguments from different `setEnv` calls ourselves and call the `EnvironmentPlugin` only once?

packages/webpack/README.md
```diff
@@ -79,7 +79,23 @@ Replaces constants in your source code with a value (`process.env.NODE_ENV` for
...
 Every constant's value is `JSON.stringify()`-ed first, so you don't have to remember.
 
-Special feature: Using `defineConstants` multiple times results in a single DefinePlugin instance configured to do all the replacements.
+Using `defineConstants` multiple times results in a single DefinePlugin instance configured to do all the replacements.
+
+#### setEnv(constants: string[]|object): Function
+
+Replaces constants in your source code with a values from `process.env` using the [webpack.EnvironmentPlugin](https://webpack.js.org/plugins/environment-plugin/).
+
+Using `setEnv` multiple times results in a single EnvironmentPlugin instance configured to do all the replacements.
```
vlad-zhukov: I've removed the 'Special feature' note here and above in `defineConstants`. The sentence feels better to me without additional emphasis.

packages/webpack/lib/setEnv.js
```diff
@@ -0,0 +1,37 @@
...
+function addEnvironmentPlugin (context, util) {
+  return util.addPlugin(
+    new context.webpack.EnvironmentPlugin(context.setEnv)
+  )
+}
+
+function toObject (constants) {
+  if (Array.isArray(constants)) {
+    return constants.reduce((acc, c) => {
+      acc[c] = null
```
vlad-zhukov: Setting to `null` according to [EnvironmentPlugin docs](https://webpack.js.org/plugins/environment-plugin/).

> To specify an unset default value, use `null` instead of `undefined`.

packages/webpack/lib/setEnv.js
andywer: I'd say yes. Seems cleaner. Not sure if multiple plugin instances have side effects.

packages/webpack/README.md
sapegin: 👍

packages/webpack/lib/setEnv.js
sapegin: Maybe we should add this note as a comment.

packages/webpack/lib/setEnv.js
sapegin: Cleaner to who? ;-) That would significantly complicate the code. The only side effect I can imagine is performance.

andywer: Yeah, if it works seemlessly with multiple plugin instances and changing it would make the code noticeably more complex, then nevermind.
No need to make our lives unnecessarily hard 😉

packages/webpack/lib/setEnv.js
vlad-zhukov: Setting to `null` silences [this warning](https://github.com/webpack/webpack/blob/master/lib/EnvironmentPlugin.js#L30). Not sure if it's good.

vlad-zhukov: @andywer @sapegin WDYT? It should be `undefined` instead of `null` imo.

andywer: Shouldn't it rather be `acc[c] = process.env[c]`?

andywer: Btw, `acc` and `c` are not very good identifiers 😉

vlad-zhukov: The [EnvironmentPlugin does it](https://github.com/webpack/webpack/blob/master/lib/EnvironmentPlugin.js#L26) so it makes no sense to do that.

`acc` is for accumulator and kind of standard, `c` is easy to understand from the code. But no problem, I will replace them with `result` and `constant`.

sapegin: > Shouldn't it rather be acc[c] = process.env[c]?

We’re setting default values here, and EnvironmentPlugin will use them these keys are not defined in `process.env`. It’s a bit hard to understand from their docs ;-)

packages/webpack/README.md
```diff
@@ -79,7 +79,23 @@ Replaces constants in your source code with a value (`process.env.NODE_ENV` for
...
+
+```js
+module.exports = createConfig([
+  setEnv(['NODE_ENV']),
+  setEnv({
+    BABEL_ENV: 'development',
```
sapegin: We should also explain that these are only default values.

vlad-zhukov:
Updated and rebased

packages/webpack/lib/setEnv.js
```diff
@@ -0,0 +1,37 @@
...
+function addEnvironmentPlugin (context, util) {
+  return util.addPlugin(
+    new context.webpack.EnvironmentPlugin(context.setEnv)
+  )
+}
+
+function toObject (constants) {
+  if (Array.isArray(constants)) {
+    return constants.reduce((result, constant) => {
+      result[constant] = undefined
```
sapegin: I think this place needs an explanation. It’s not clear what we need this unless you know how EnvironmentPlugin works.

andywer: Don't know how you feel about it, but maybe we shouldn't rely on that implicit magic and rather go for an explicit default here in the `setEnv()` block. Also it looks more like a hack than a documented feature, so I would fear that this behavior might break with a future update.

andywer:
We should have a look if the `undefined` magic is a good idea at all (see my comment). If it is we should add at least a comment. Otherwise fine to me as well :)

packages/webpack/lib/setEnv.js
sapegin: It is documented an we’ve discussed that already ;-)

But it supposed to be `null` as I understand.

vlad-zhukov: There is nothing magical in here, EnvironmentPlugin resolves arrays in a very similar fashion. The reason it's `undefined` is to have the following warning:
```
WARNING in EnvironmentPlugin - BABEL_ENV environment variable is undefined.

You can pass an object with default values to suppress this warning.
See https://webpack.js.org/plugins/environment-plugin for example.
```
Defaulting to `null` will swallow the warning which is not useful.

andywer: Ahhh, I think I just got the idea. We normalize arrays to objects to make it all composable. Setting the default value to `undefined` achieves that, setting to `null` does not, since the `EnvironmentPlugin` would think that `null` is the intended default value.
Right?

packages/webpack/lib/setEnv.js
```diff
@@ -0,0 +1,40 @@
...
+function addEnvironmentPlugin (context, util) {
+  return util.addPlugin(
+    new context.webpack.EnvironmentPlugin(context.setEnv)
+  )
+}
+
+function toObject (constants) {
+  if (Array.isArray(constants)) {
+    return constants.reduce((result, constant) => {
+      // Setting the default (fallback) value to `undefined`
```
andywer: 👍

vlad-zhukov: The thing is all these `undefined`s and `null`s are resolved in DefinePlugin, and yes, it lacks documentation like many other things, but I don't believe it's an excuse to introduce a breaking change in a minor release for them.

sapegin:
Let’s merge it ;-)

andywer:
Just added one commit updating the CHANGELOG.

@sapegin Do you have permissions to merge PRs? 😉

sapegin:
Yup, let’s just wait for the CI ;-)

sapegin:
Merged, thanks @vlad-zhukov!

PR #213, opened 2017-09-18
Add minimize option to postcss and sass blocks

sapegin:
For consistency with the `css` block.

### DIFF
--- a/packages/postcss/CHANGELOG.md
+++ b/packages/postcss/CHANGELOG.md
@@ -1,6 +1,6 @@
 # @webpack-blocks/postcss - Changelog
 
-## 1.0.0-beta.5
+## Next release
 
 - Add `minimize` option.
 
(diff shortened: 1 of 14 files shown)


packages/postcss/README.md
```diff
@@ -41,6 +41,9 @@ module.exports = {
 
 ## Options
 
+#### minimize *(optional)*
+Enable CSS minification.
```
vlad-zhukov: I think it should say that it's handled by the `css-loader`.

vlad-zhukov: Also need to add this options to sass block readme as well.

sapegin: Good points — done ;-)

packages/postcss/CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # @webpack-blocks/postcss - Changelog
 
+## 1.0.0-beta.5
```
andywer: `Next release` might be better. The next release could just as well be `1.0.0-rc`.

PR #216, opened 2017-09-18
Add tests to test-app

### LINKED ISSUE #66: Add simple testing to test-app
I fear we will always have issues with the test-app unless there is some basic testing, even though the test app is no production code.

tdeschryver:
Resolves #66.


test-app/__tests__/webpack.config.babel.js
```diff
@@ -0,0 +1,11 @@
+import test from 'ava'
+import config from '../webpack.config.babel'
+
+test('it creates the config file', t => {
+  const plugins = config.plugins
+  delete config.plugins // because it has absolute paths and bring a lot of bloat
```
tdeschryver: Is this OK?
Or do you know solution for this, if not I'll take a deeper look.

sapegin: Probably we can do another check for plugins, at least that all plugins are present.

test-app/__tests__/webpack.config.babel.js
```diff
@@ -0,0 +1,11 @@
+import test from 'ava'
+import config from '../webpack.config.babel'
+
+test('it creates the config file', t => {
```
sapegin: I think it exports configuration, not creates a file.

test-app/__tests__/webpack.config.babel.js
```diff
@@ -0,0 +1,11 @@
...
+  const plugins = config.plugins
+  delete config.plugins // because it has absolute paths and bring a lot of bloat
+
+  t.snapshot(config)
+
+  config.plugins = plugins
```
sapegin: I’d avoid mutations here: `_.omit(config, 'plugins')`.

test-app/__tests__/webpack.config.babel.js
tdeschryver: I think I was a bit too eager to send a PR yesterday.
After giving this a second look I believe that a snapshot test might not be needed because: 
- each block should have its own separate test
- the `createConfig` has its own test

So maybe a sanity check to check it isn't throwing any errors should be enough?

andywer:
Should we close the PR for now?

sapegin:
@andywer Why do you want to close it? ;-) I think we should merge it as soon as conflicts are resolved.

andywer:
[This comment](https://github.com/andywer/webpack-blocks/pull/216#discussion_r139753479) sounded like there are rather fundamental concerns. But maybe I am mistaken there 😉

tdeschryver:
@andywer the test is OK for me.
The thing about the snapshot testing vs a sanity check is just a matter of preference I think and should be OK for now?

andywer:
We should probably add a `posttest` script to the top-level `package.json`, running `cd test-app/ && npm test`. I can quickly add it as well, just asking for opinions first 😉

Btw, this is a good first step, but an end-to-end test (checking the build files) would increase the confidence of this test quite a bit. Should consider that in the future.

tdeschryver:
@andywer OK if I wait a few days before adding the `posttest`, so we can hear some opinions first? 
I'm in favor of adding it btw!
Ill also try to add a end-to-end test in during the week, would it be enough to check if there are files in the build folder?

sapegin:
`cd test-app && npm i && npm t`? ;-)

test-app/__tests__/webpack.config.babel.js
```diff
@@ -0,0 +1,29 @@
+import test from 'ava'
```
andywer: This test file shouldn't be called `webpack.config.babel.js`. How about `webpack-config.test.js`?

test-app/__tests__/webpack.config.babel.js
```diff
@@ -0,0 +1,29 @@
...
+  t.snapshot(config)
+})
+
+test.cb('it builds', t => {
+  exec('npm run build', (err, stdout, stderr) => {
+    t.is(err, null)
+
+    const html = fs.readFileSync('./build/index.html', { encoding: 'utf8' })
+    const js = fs.readFileSync('./build/bundle.js', { encoding: 'utf8' })
+    const css = fs.readFileSync('./build/css/main.0ec7a2a4.css', { encoding: 'utf8' })
```
andywer: Doesn't the nonce in the filename change on every build?

test-app/__tests__/webpack.config.babel.js
tdeschryver: Yep good for me, is indeed better.

test-app/__tests__/webpack.config.babel.js
tdeschryver: It will only change when something from the css changes (same for the class name). 
Ill push a new commit soon, which handles it a bit differently - thoughts welcome!

test-app/__tests__/webpack-config.test.js
```diff
@@ -0,0 +1,87 @@
...
+
+test('it exports the production config', t => {
+  const { NODE_ENV } = process.env
+  process.env.NODE_ENV = 'production'
+
+  delete require.cache['../webpack.config.babel.js']
+  const originalConfig = requireConfig()
+  const module = Object.assign({}, originalConfig.module, {
+    rules: originalConfig.module.rules.map((rule, index) => {
+      if (index === 2) {
```
sapegin: This doesn‘t look good. Could we use something else instead of an index?

sapegin: Yeah, may work ;-)

sapegin:
@andywer @tdeschryver What’s let here?

tdeschryver:
For me its good to go (just rebased to resolve conflicts).

andywer:
Looks good to me! Will need some e2e testing later.

@sapegin How about you?

sapegin:
Merged ;-)

PR #218, opened 2017-09-19
PostCSS: Remove plugins argument

sapegin:
**This is a breaking change**

Also do not use LoaderOptionsPlugin and remove a hack for sass-loader that is no longer relevant.

### DIFF
--- a/packages/postcss/index.js
+++ b/packages/postcss/index.js
@@ -9,15 +9,19 @@ const _ = require('lodash')
 module.exports = postcss
 
 /**
- * @param {PostCSSPlugin[]} [plugins]               Will read `postcss.config.js` file if not supplied.
  * @param {object}          [options]               https://github.com/postcss/postcss-loader#options
  * @param {bool}            [options.minimize]      Enable minification.
  * @param {string}          [options.parser]        Package name of custom PostCSS parser to use.
  * @param {string}          [options.stringifier]   Package name of custom PostCSS stringifier to use.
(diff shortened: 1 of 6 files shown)


packages/postcss/index.js
```diff
@@ -9,15 +9,19 @@ const _ = require('lodash')
...
  * @param {string}          [options.parser]        Package name of custom PostCSS parser to use.
  * @param {string}          [options.stringifier]   Package name of custom PostCSS stringifier to use.
  * @param {string}          [options.syntax]        Package name of custom PostCSS parser/stringifier to use.
+ * @param {PostCSSPlugin[]} [options.plugins]       PostCSS plugins, will read `postcss.config.js` file if not supplied.
  * @return {Function}
  */
-function postcss (plugins = [], options = {}) {
+function postcss (options = {}) {
+  if (Array.isArray(options)) {
+    throw Error('Passing PostCSS plugins as a first argument is not supported anymore, use options.plugins instead')
```
andywer: 👍

packages/postcss/index.js
```diff
@@ -37,23 +41,6 @@ function postcss (plugins = [], options = {}) {
...
     return nextConfig
   }
 }
-
-function createLoaderOptionsPlugin ({ webpack }, postcssPlugins) {
-  return new webpack.LoaderOptionsPlugin({
-    options: {
-      postcss: postcssPlugins,
-
-      // Hacky fix for a strange issue involving the postcss-loader, sass-loader and webpack@2
```
andywer: 👍

PR #219, opened 2017-09-20
Refactor webpack-blocks/core

vlad-zhukov:
Currently the index file contains all of the main code, it's quite large and way too hard to understand what's going on in there. Also it's difficult or even impossible to test helper functions (such as hooks).

In this PR I split the index file into multiple files and also replaced 2 small helper functions &ndash; `isFunction` and `filterDuplicates` &ndash; with ones from `lodash`. Didn't do any code changes apart from those.


packages/core/lib/createConfig.js
```diff
@@ -0,0 +1,53 @@
...
+  }
+  assertConfigSetters(configSetters)
+
+    /** @deprecated context.fileType */
+  const fileType = createFileTypesMapping(defaultFileTypes)
+  const context = Object.assign({ fileType }, initialContext)
+
+  const baseConfig = {
+    resolve: {
+            // Explicitly define default extensions, otherwise blocks will overwrite them instead of extending
```
sapegin: Looks like something is wrong with indentation here and in the next comment.

packages/core/lib/createConfig.js
```diff
@@ -0,0 +1,53 @@
+const createFileTypesMapping = require('./createFileTypesMapping')
+const defaultFileTypes = require('./defaultFileTypes')
+const {assertConfigSetters, invokeConfigSetters} = require('./configSetters')
+const {invokePreHooks, invokePostHooks} = require('./hooks')
```
sapegin: We used to have spaces inside {}, not sure if there are any linting rules for that.

vlad-zhukov: `standard` didn't catch these, I'll fix manually then.

sapegin: I’m trying to sell Prettier to @andywer — but after 1.0 final release ;-)

andywer:
Great work, Vlad! Thank you 😊

And thx for the review, Artem. I agree. Good points!

packages/core/lib/createConfig.js
vlad-zhukov: My IDE formatted the code on insertion, but again `standard` neither caught nor fixed those 😞 . I fixed few cases manually, but missed those.

It would be great to have a top-level prettier script and stop wasting time on formatting.

sapegin: 👍

vlad-zhukov:
Fixed formatting issues

PR #220, opened 2017-09-20
Test app cross env

### LINKED ISSUE #215: test-app on Windows
The test-app is throwing errors when building or starting on windows because of `NODE_ENV=xxx`

Proposal:
Using [kentcdodds/cross-env](https://github.com/kentcdodds/cross-env) so we can do `cross-env NODE_ENV=xxx`

tdeschryver:
Adding `cross-env` to make the test app cross platform.
Closes #215.

### DIFF
--- a/test-app/package.json
+++ b/test-app/package.json
@@ -1,7 +1,7 @@
 {
   "scripts": {
-    "build": "NODE_ENV=production webpack --config webpack.config.babel.js",
-    "start": "NODE_ENV=development webpack-dev-server --config webpack.config.babel.js"
+    "build": "cross-env NODE_ENV=production webpack --config webpack.config.babel.js",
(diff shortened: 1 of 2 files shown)


test-app/package.json
```diff
@@ -18,5 +18,8 @@
     "webpack": "^3.5.3",
     "webpack-blocks": "^1.0.0-beta.4",
     "webpack-dev-server": "^2.7.1"
+  },
+  "dependencies": {
+    "cross-env": "^5.0.5"
```
vlad-zhukov: It should be a dev dependency

sapegin: 👍

tdeschryver: Oh you're right, thanks!

PR #223, opened 2017-09-22
Add v0.4 to v1.0 migration guide

andywer:


docs/MIGRATION-GUIDE.md
```diff
@@ -0,0 +1,208 @@
...
+
+The blocks still work without `match()`. **However, blocks dropped the `fileType`, `exclude` & `include` from their options. Use `match()` for that.**
+
+```js
+const { createConfig, css, file, match, postcss, url } = require('webpack-blocks')
+const path = require('path')
+
+module.exports = createConfig([
+  babel(),            // matches *.js, *.jsx outside node_modules/ by default
+  match('*.css', { exclude: path.resolve('node_modules') }, [
```
vlad-zhukov: Weird comment formatting, also why use `path.resolve` when a regex `/node_modules/` works fine? The main readme uses `path.resolve` as well...

andywer: Yeah, I just copied the snippet. Where would you put the comment?

`path.resolve()` vs regex seems almost like a philosophical question. One is precise, but longer, the other one is concise, but inprecise, yet will do the job as well.

vlad-zhukov: I'd put it right next to the block, and would also add new lines between blocks for better readability.

andywer: Isn't it next to the block? (Maybe you can just paste an example of what you suggest)

vlad-zhukov: There is a huge whitespace between
screen shot 2017-09-22 at 13 42 39

andywer: Now we're talking. How about this?

```js
const { createConfig, css, file, match, postcss, url } = require('webpack-blocks')
const path = require('path')

  babel(),    // matches *.js, *.jsx outside node_modules/ by default

  match('*.css', { exclude: path.resolve('node_modules') }, [
    css(),
    postcss()
  ]),

  match(['*.gif', '*.jpg', '*.jpeg', '*.png', '*.svg', '*.webp'], [
    file()
  ])
])
```

sapegin: I’d put it on top of the block, and capitalize the first letter.

README.md
```diff
@@ -43,6 +44,11 @@ yarn add --dev webpack webpack-blocks
 ```
 
 
+## Upgrade from v0.4
+
+Do you have an existing webpack-blocks v0.4 setup? Check out our [migration guide](./docs/MIGRATION-GUIDE.md) to get started with v1.0 today.
```
sapegin: “Do you have an existing webpack-blocks v0.4 setup?” → “”.

docs/MIGRATION-GUIDE.md
```diff
@@ -0,0 +1,212 @@
...
+- [Migrating blocks](#migrating-blocks)
+
+<!-- tocstop -->
+
+
+## Migrating configuration
+
+### `match()`
+
+You can now use `match()` to specify on which files to apply certain loaders. Works with every block that adds a loader, like `babel`, `css`, `elm`, `postcss`, `sass`, ...
```
sapegin: No need for ellipsis, you already have “like”.

docs/MIGRATION-GUIDE.md
```diff
@@ -0,0 +1,212 @@
...
+    file()
+  ])
+])
+```
+
+### `@webpack-blocks/assets` package
+
+The `assets` package contains the `css`, `css.modules`, `file` and `url` blocks which were contained in `@webpack-blocks/webpack` in v0.4. Use these blocks to configure a `css-loader`, `file-loader` or `url-loader`.
+
+See the package's README for further details.
```
sapegin: Link?

docs/MIGRATION-GUIDE.md
```diff
@@ -0,0 +1,212 @@
...
+* babel
+* dev-server
+* extract-text
+* postcss
+* sass
+* typescript
+* uglify
+* webpack
+
+It also exports all the `@webpack-blocks/webpack` utility functions. See the package's README for further details.
```
sapegin: Link?

docs/MIGRATION-GUIDE.md
```diff
@@ -0,0 +1,212 @@
...
+
+### No more webpack v1
+
+Support for webpack 1.x has been dropped. If you have not upgraded yet we recommend doing so soon.
+
+### webpack as a peer dependency
+
+Webpack is now a peer dependency of `@webpack-blocks/webpack`. Just make sure to explicitly `npm install --save-dev webpack`.
```
sapegin: “Just” → “”.

docs/MIGRATION-GUIDE.md
```diff
@@ -0,0 +1,212 @@
...
+
+The `@webpack-blocks/babel6` package has been renamed to `@webpack-blocks/babel`. You can now choose which version of babel to use, since `babel-core` is now a peer dependency (see below).
+
+### babel-core as a peer dependency
+
+`babel-core` is now a peer dependency of the `babel` block, so you can control which version of `babel-core` to use. Make sure to `npm install --save-dev babel-core` in your application.
+
+### `webpack` no longer exported
+
+The `webpack` instance is no longer exported by `@webpack-blocks/webpack`. Use `const webpack = require('webpack')` instead which is more explicit.
```
sapegin: I think the term “instance” is used incorrectly here because it’s a reexport of the whole `webpack` module.

docs/MIGRATION-GUIDE.md
```diff
@@ -0,0 +1,212 @@
...
+
+### webpack v3
+
+We officially support webpack 3 now 🎉
+
+### `@webpack-blocks/uglify` package
+
+The `uglify` package is a convenience block for easily configuring JS minification using the [uglifyjs-webpack-plugin](https://www.npmjs.com/package/uglifyjs-webpack-plugin) which also supports minifying untranspiled ES2015 code.
+
+See the package's README for further details.
```
sapegin: Link?

docs/MIGRATION-GUIDE.md
```diff
@@ -0,0 +1,212 @@
...
+
+This way extensions you set later in your `createConfig()` will have precedence over extensions you set earlier (webpack tries to match against the `extensions` from the first to last).
+
+---
+
+## Migrating blocks
+
+### A concise migration example
+
+A simple babel block in the old v0.4:
```
sapegin: Probably not the best example — it feels like I have to do something with the default `babel` block.

docs/MIGRATION-GUIDE.md
```diff
@@ -0,0 +1,212 @@
...
+
+
+### New API (#125)
+
+In webpack-blocks v0.x a block was a function with the signature `(context) => configDiff`.
+
+In webpack-blocks v1.0 a block is now expected to have the signature
+
+```
+(context, utils) => prevConfig => updatedConfig
```
sapegin: Diff?

```diff
- (context) => configDiff
+ (context, utils) => prevConfig => updatedConfig
```

docs/MIGRATION-GUIDE.md
```diff
@@ -0,0 +1,212 @@
...
+])
+```
+
+### Removed `devServer.reactHot()`
+
+The `devServer.reactHot()` block has never worked too well. We also decided not to ship framework-specific blocks as core features.
+
+If you feel like this is a feature you really want to have, create a new fancy React hot loader block and share it with the world!
+
+### Requiring node 6+ now
```
sapegin: node → Node.js.

docs/MIGRATION-GUIDE.md
andywer: Yeah, but it's easy to understand and nearly everyone knows it. Couldn't think of a better example.

andywer:
Updated.

andywer:
Updated everything, except the `babel` block sample, since I can't think of a better one right now. 👍 / 👎 ?

docs/MIGRATION-GUIDE.md
```diff
@@ -0,0 +1,214 @@
...
+
+### `resolve()` will prepend custom `extensions` (instead of appending them)
+
+The new `resolve` utility in `@webpack-blocks/webpack` will prepend custom `extensions` you pass to it, rather than appending them.
+
+This way extensions you set later in your `createConfig()` will have precedence over extensions you set earlier (webpack tries to match against the `extensions` from the first to last).
+
+---
+
+## Migrating blocks
```
vlad-zhukov: `Migrating custom blocks`?

docs/MIGRATION-GUIDE.md
vlad-zhukov: I think there might not be enough context for people to understand what it is for. There should be an introduction paragraph like `In version 1.0 we are introducing a new API for writing blocks`, and also probably a few words to answer a question 'Why?' (to conform with the new block usage, to make an expected effect of blocks more explicit, ...).

Then I'd use something like
```
A simple Babel block written for the old v0.4 API would look like this:
```
And later
```
The same block updated for the v1.0 API:
```

docs/MIGRATION-GUIDE.md
andywer: Good point! Changed it.

docs/MIGRATION-GUIDE.md
andywer: I added an introductary paragraph to the top of the migration guide. Does that work for you or should I change it?

Also updated the captions for the sample babel block.

sapegin:
👍

andywer:
Thx for reviewing, guys! :)

PR #224, opened 2017-09-22
CSS: Allow to pass options to style-loader and allow to remove it from loaders

vlad-zhukov:
~WIP because I haven't updated docs yet.~ Just wanted to get some feedback.

Changes:
- added `styleLoader` option to both `css` and `css.modules` blocks
- if it's `false` a `style-loader` won't be added
- added tests for `styleLoader` and `css.modules`

### DIFF
--- a/packages/assets/__tests__/css-modules.test.js
+++ b/packages/assets/__tests__/css-modules.test.js
@@ -2,9 +2,9 @@ import test from 'ava'
 import { createConfig, match } from '@webpack-blocks/core'
 import { css } from '../lib/index'
 
-test('css() works', t => {
+test('css.modules() works', t => {
   const config = createConfig({}, [
-    css()
+    css.modules()
   ])
 
   t.deepEqual(config.module.rules, [
@@ -16,17 +16,21 @@ test('css() works', t => {
           options: {}
         }, {
           loader: 'css-loader',
-          options: {}
+          options: {
+            importLoaders: 1,
+            localIdentName: '[name]--[local]--[hash:base64:5]',
+            modules: true
+          }
         }
       ]
     }
   ])
 })
 
-test('css() works with options and match()', t => {
+test('css.modules() works with options and match()', t => {
   const config = createConfig({}, [
     match('*.pcss', { exclude: /node_modules/ }, [
-      css({
+      css.modules({
         minimize: true
       })
     ])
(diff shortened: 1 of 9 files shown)


packages/assets/__tests__/css-modules.test.js
```diff
@@ -0,0 +1,164 @@
...
+            localIdentName: '[name]--[local]--[hash:base64:5]',
+            modules: true
+          }
+        }
+      ]
+    }
+  ])
+})
+
+test('deprecated css(<fileType>) still works', t => {
```
andywer: I wouldn't repeat all those tests for `css.modules()`. I think it's ok to only test those backwards-compatibility stuff in one of the blocks (`css()`).

packages/assets/lib/css.js
```diff
@@ -51,22 +57,23 @@ function cssModules (fileType, options = {}) {
...
     importLoaders: 1,
     localIdentName: String(process.env.NODE_ENV) === 'production'
-      ? '[hash:base64:10]'
-      : '[name]--[local]--[hash:base64:5]'
-  }, options)
+            ? '[hash:base64:10]'
+            : '[name]--[local]--[hash:base64:5]'
+  }, _.omit(options, ['exclude', 'include', 'styleLoader']))
+  const loaders = [{ loader: 'css-loader', options: cssLoader }]
+  if (options.styleLoader !== false) {
```
andywer: Is this breaking change intended? Now you actually need to explicitly set `options.styleLoader` to get a `style-loader`.

I would think a vast majority of the users would like to have the style-loader by default and opt-out if necessary.

andywer: Where do I find that in the code? Don't see it...

sapegin: By default it’s `undefined`, which is !== `false` ;-)

andywer: Right. It's getting late... ;P

Just wondering if we can somehow make the code a bit more speaking in this particular place.

vlad-zhukov:
Fixed nits and updated docs.

sapegin:
Looks like you need to update snapshots.

vlad-zhukov:
CI is green

andywer:
@vlad-zhukov Please don't squash the commits before pushing. We have to review everything from scratch after each push... 🙈😉

andywer:
But thx for fixing

andywer:
Fine for me. @sapegin Any objections?

sapegin:
Merged :shipit:

vlad-zhukov:
Thanks! 🎉

PR #226, opened 2017-09-25
Improve build

andywer:
Speed-up build (almost one third faster).

Changes:
- [x] Install `ava` & `standard` monorepo-wide, not in each package
- [x] Rename `prepublish` scripts to `prepublishOnly`, so the tests are not run at `npm install` (read more [here](https://resin.io/blog/safely-migrating-away-from-prepublish-with-npm-4/))

bildschirmfoto 2017-09-25 um 22 39 03


sapegin:
Why do we need link-parent-bin?

andywer:
@sapegin Because otherwise there will be no `packages/*/node_modules/.bin/{ava,standard}` if `ava` and `standard` are only installed in the parent package 😉

sapegin:
You need to remove `test` script from each package and probably setup paths. That should make it even faster.

See an example here: https://github.com/sapegin/mrm-tasks

andywer:
We can try and benchmark, but it shouldn't be substantially faster, I think. (But maybe you have tried already?)

What do you mean by "setup paths"?

sapegin:
I don’t think we need to benchmark it. Lerna will just run ava and standard once for the whole repo. 

> What do you mean by "setup paths"?

I guess the defaults should work but you may need to [ignore some folders](https://github.com/sapegin/mrm-tasks/blob/master/.eslintignore).

andywer:
Actually I like being able to run `npm test` in a single child package... And I would not want to duplicate the test script between root package.json and child packages.

Wouldn't touch that for now.

sapegin:
Watch mode should work equally fine and the setup it 10 times easier ;-)

package.json
```diff
@@ -1,9 +1,12 @@
 {
   "scripts": {
     "test": "lerna run test",
-    "postinstall": "lerna bootstrap"
+    "postinstall": "lerna bootstrap && link-parent-bin"
   },
   "devDependencies": {
-    "lerna": "2.0.0-beta.30"
+    "ava": "^0.22.0",
+    "lerna": "2.0.0-beta.30",
```
vlad-zhukov: Mind upgrading `lerna` to the latest?

andywer: I think there was an issue with newer versions that prevented me from upgrading in the past. But yeah, we should have a look at that 😉

andywer:
@sapegin I tried invoking `ava` and `standard` only once at top-level, but there is an issue with the end-to-end tests. Webpack using the node API fails, because it cannot find the loader packages, since cwd is the top-level directory, but the loaders are installed in `/packages/webpack`.

Got to work around that somehow.

andywer:
How about merging this and maybe open a follow-up PR later? These changes here already improve build times noticeably.

PR #114, opened 2017-02-15
Added less block

Veikedo:
Added less block.
@kirill-konshin please test the block since I can't do it for a while.

PR #229, opened 2017-10-08
Add webpack-blocks-less

sapegin:
Close #114.


andywer:
Good catch, but I think we only added the LESS block to the 0.x branch's list, because it's not yet compatible with v1.0.

Opened an issue for that: kirill-konshin/webpack-blocks-less#3

PR #230, opened 2017-10-09
Add inspect method

### LINKED ISSUE #200: Debug utility
It’s often useful to see what config is generated by webpack-blocks and installing stringify-object just for that isn’t that convenient. It would be cool to have an internal debug utility.

Possible API:

```js
  /* ... */
]).debug();
```

```js
  /* ... */
]).debug('module.rules');
```

`debug` method will nicely print the config (or it’s part) and return the config back so it would still work.

What do you think?

---

I suggest continue using [stringify-object](https://github.com/yeoman/stringify-object) for printing and [Lodash’s get](https://lodash.com/docs/4.17.4#get) method to access deeply nested objects inside the config.

Please send your pull request to `release/1.0` branch.

sapegin:
Closes #200


README.md
```diff
@@ -238,16 +239,17 @@ Missing something? Write and publish your own webpack blocks!
...
-const stringify = require('stringify-object')
+const { createConfig, inspect } = require('webpack-blocks')
 
 module.exports = createConfig([
   // ...
 ])
 
-console.log(stringify(module.exports))
+inspect(module.exports) // Print the whole config
+inspect(module.exports, 'module.rules') // Print part of the config
```
vlad-zhukov: Why would I want to do `inspect(module.exports, 'module.rules')` instead of `inspect(module.exports.module.rules)`? The only reason I can think of is if it throws a more meaningful error rather than the default one if `modules` is `undefined` or `null` for example.

sapegin: You don’t have to do it ;-)

vlad-zhukov: [It will return `undefined`](https://github.com/developit/dlv) if path doesn't exist.

sapegin: You can use `inspect(module.exports.module.rules)` if you like ;-)

packages/webpack/README.md
```diff
@@ -59,6 +59,20 @@ module.exports = createConfig([
...
+
+Prints an object with color highlighting for easier debugging.
+
+Use like this:
+
+```js
+module.exports = createConfig([
+  // ...
+])
+inspect(module.exports) // Print the whole config
```
andywer: There should be a `const inspect = require('...')` in the snippet 😉

packages/core/lib/inspect.js
```diff
@@ -0,0 +1 @@
+module.exports = require('q-i').print
```
andywer: 1. `@webpack-blocks/webpack` does not actually require or export `inspect()`.
2. Why not move the `inspect()` code to `@webpack-blocks/webpack` directly? It's not really core stuff, I'd say.

packages/webpack/README.md
sapegin: Then there should be `require` for `createConfig` in all examples.

packages/core/lib/inspect.js
sapegin: I can move it to `webpack` — I don’t see any difference between webpack and core.

README.md
vlad-zhukov: Sure but then if `module` is `undefined` it will throw. 👆

packages/webpack/README.md
andywer: Ahh, forget it. The snippet is in the list of exports, so it should be clear.

packages/core/lib/inspect.js
andywer: I think `@webpack-blocks/webpack` feels more natural, since it's utility functionality, not core functionality.

sapegin:
I’ve moved `inspect` to the `webpack` package and have added an export.

andywer:
Could you please update the `@webpack-blocks/webpack` CHANGELOG instead of the core one? I think it should be fine then.

The build is failing right now, but it seems to be because of the uglify plugin issue...

andywer:
I think @sapegin is still on vacation.

Something to think about when you are back: I find the `q-i` package really useful for debugging the webpack config. I just started to think if we should actually package it as part of webpack-blocks, since it does not provide a benefit over installing `q-i` separately.

Maybe it's better to keep webpack-blocks dependencies as slim as possible (it's already pretty heavy) and just make `q-i` the official way to debug webpack-blocks configs in the docs.

sapegin:
It’s super tiny, I wouldn’t worry about that ;-)

P. S. I’m between two trips and my train was cancelled today, will try to leave Berlin again tomorrow ;-|

andywer:
Ok :) Couldn't find a working site that can tell me the size of a bundled and minified `q-i` yet 🙃

Yeah, it's not the best time to go by train, I guess 😉

vlad-zhukov:
@andywer I am wondering why do you care of the size of `q-i`? `webpack-blocks` is mostly a dev dependency... I am honestly interested what's the use-case why it matters.

andywer:
@vlad-zhukov I am not too concerned about the size, but of course every additional dependency will slow down the `npm install` a little bit more, and it might not be necessary.

What is more interesting to me right now is actually if we should ship `q-i` with `webpack-blocks`, since the `inspect()` function is just a proxy for `q-i`. There seems to be no benefit over just installing `q-i` alongside webpack-blocks.

But we can see if that's true when Artem is back.

sapegin:
Maybe we shouldn’t I’m just to lazy to type `npm i --no-save q-i` every time ;-) But you’re probably right since we couldn’t implement my initial API idea, where you won’t have require the method separately.

PR #231, opened 2017-10-10
Fix validation error with default uglify options

jvanbruegge:
When running a webpack config with the default options the build fails due to a validation error.


sapegin:
Did they change something? 

https://github.com/webpack-contrib/uglifyjs-webpack-plugin#parallel

jvanbruegge:
I dont think so, but the validation is not throwing an error

sapegin:
Then if boolean for `parallel` option is incorrect we should send a PR to uglifyjs-webpack-plugin too.

sapegin:
Yup, it should work:

https://github.com/webpack-contrib/uglifyjs-webpack-plugin/blob/ee1663991b37f243a65d0ee69e37169e6e113943/src/uglify/index.js#L19-L21

sapegin:
https://github.com/webpack-contrib/uglifyjs-webpack-plugin/issues/139

sapegin:
They’ve made a breaking change without any change logs, the correct settings are:

```
parallel: true,
cache: true,
```

sapegin:
We should also lock the uglifyjs-webpack-plugin version in package.json.

andywer:
Thx for investigating, @sapegin! 👍

sapegin:
Now the question is: why the CI is failing?

sapegin:
@andywer Any ideas?

jvanbruegge:
No idea

andywer:
@sapegin Don't know. Just restarted the failed build. Maybe some issue with node 6 vs. node 8.

andywer:
An end-to-end test fails and maybe lerna swallows the error message.

vlad-zhukov:
v1.0 was released

andywer:
@vlad-zhukov Thx for the hint! 🙂

andywer:
@jvanbruegge Can we just merge jvanbruegge/webpack-blocks#1? Then we should be able to merge this PR, release a new version of the uglify block and the other PT builds should become green as well :)

Would just be good if someone would have a look at my changes in jvanbruegge/webpack-blocks#1

jvanbruegge:
Done

andywer:
Will merge and release this evening :)

andywer:
@vlad-zhukov 
>  I'd put the sample app into an examples folder and add it to lerna

How would you add it to lerna?

vlad-zhukov:
@andywer [using the `packages` field](https://github.com/lerna/lerna#lernajson). I can submit a PR tomorrow.

andywer:
@vlad-zhukov Ahh, alright. Good idea 👍

andywer:
Published as `@webpack-blocks/uglify v1.1.0` & `webpack-blocks v1.0.0-rc.2` 🚀

PR #236, opened 2017-10-30
Upgrade lerna setup

### LINKED ISSUE #204: Try lerna hoisting
Try [lerna bootstrap --hoist](https://github.com/lerna/lerna/blob/master/doc/hoist.md) to improve installation time for dev set-up.

vlad-zhukov:
I made a few changes we've discussed in #226, #231, #204
Closes #204

What have I done:
- upgraded `lerna` to 2.4.0 (https://github.com/andywer/webpack-blocks/pull/226#discussion_r141167531)
- added `--hoist` flag to `lerna bootstrap` (#204) and ran it with the latest `npm@5.5.1` (it's responsible for upgrading all deps and it's not possible to revert it, we have to use `yarn` if we want more predictable behaviour in this department)
- moved the `sample-app` into `examples` folder (https://github.com/andywer/webpack-blocks/pull/231#pullrequestreview-72118837)
- added `--parallel` flag to `lerna run test`, it should be faster now (we haven't discussed it before but I find it interesting to try)
- tried to fix a skipped test in the `sample-app`

This is a WIP because I can't make tests pass locally due to a weird error in the `elm` block. Would be great if someone could help me with it because I've no experience with `elm`.

P.S.: Sorry for a large PR, I can try to split it if you want.


examples/sample-app/__tests__/webpack-config.test.js
```diff
@@ -46,16 +46,16 @@ test('it exports the production config', t => {
   process.env.NODE_ENV = NODE_ENV
 })
 
-test.skip('it builds', t => {
+test.only('it builds', async t => {
```
vlad-zhukov: Forgot to remove it, will fix later

package.json
```diff
@@ -1,11 +1,12 @@
 {
+  "private": true,
```
vlad-zhukov: Should I keep it?

examples/sample-app/__tests__/webpack-config.test.js
```diff
@@ -46,16 +46,16 @@ test('it exports the production config', t => {
...
-    testJs()
-    t.end()
-  })
+
+  await execa('npm', ['run', 'build'])
+
+  testHtml()
+  testCss()
+  testJs()
+  t.end()
```
vlad-zhukov: Tests are failing because `t.end()` is not needed.

andywer:
Thx for your PR, Vlad. Good ideas!

I'd just like to move updating the dependencies and refactoring the ugly callback-based test into separate PRs.

vlad-zhukov:
Merged with master, updated lerna to latest, bumped a few deps and fixed tests.

We have to update certain deps in this PR because of the way lerna hoists deps. For instance, if one package has a dep "webpack@^3" and another one has "webpack@^2", lerna will hoist "webpack@^3" to the root folder and won't install "webpack@^2". This is weird and I dislike such behaviour, yarn workspaces never do this.

vlad-zhukov:
Can we merge it? It's blocking me from doing more awesome stuff! 🙃

andywer:
I know that the `package-lock.json` is nasty, but it is [supposed to be committed](https://stackoverflow.com/a/44210813).

andywer:
Have been experimenting with yarn workspaces and `oao` a bit recently (see [andywer/shutter](https://github.com/andywer/shutter), for instance). Feels simpler and cleaner, actually.

Used oao instead of lerna, because lerna's "run script in all packages" output often is hardly readable.
Maybe we can discuss such an option together on gitter or in a separate issue.

vlad-zhukov:
Sure, that would be my next steps after this PR. We should also consider [`bolt`](https://github.com/boltpkg/bolt) instead of `lerna` by the way.

andywer:
Can barely tell the difference between `bolt` and `oao` ^^ But sure.

But this PR is really hard to review...

dmitmel:
@andywer @vlad-zhukov We now use Yarn workspaces, so this PR can be closed.

PR #238, opened 2017-11-01
Fix a skipped test in sample-app

vlad-zhukov:
As requested in #236 moving it in a separate PR.

### DIFF
--- a/packages/sample-app/__tests__/webpack-config.test.js
+++ b/packages/sample-app/__tests__/webpack-config.test.js
@@ -1,6 +1,6 @@
 import fs from 'fs'
 import path from 'path'
-import { exec } from 'child_process'
+import execa from 'execa'
 import rimraf from 'rimraf'
 import test from 'ava'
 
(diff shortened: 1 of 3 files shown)


andywer:
Can you please merge the current `master` into the feature branch? I hope that fixes the build.

vlad-zhukov:
I did earlier today, the typescript test is not failing anymore, but this test keep failing on Node 6 with a weird error. I couldn't reproduce it locally so far, will try look into it in the next few days.

sapegin:
Could it be because Node 6 doesn’t support async/await?

andywer:
@sapegin No, AVA transpiles test code on the fly by default.

sapegin:
Why they do it? Where they get a Babel config to do that? ;-|

sapegin:
Have found that in their docs. Another reason I like Jest more ;-£

andywer:
I really like it, actually. Provides handy defaults, but allows you to disable or customize transpiling. And you can specify different babel transforms to be applied to the test code and the tested code :)

andywer:
Closing for now.

Issue #242, opened 2017-12-06
Conditionally add blocks


boxfoot:
As I've started using webpack blocks, I've found it helpful to have a mechanism for conditionally adding blocks to the configuration -- like `env`, but keyed off any conditional, not just `NODE_ENV`.

I wrote the following block - wanted to see if this is something others are doing, and if it makes sense to include in the core:

```js
/**
 * Applies an array of webpack blocks only if the given condition is true
 * @param {boolean} condition         Condition to check
 * @param {Function[]} configSetters  Array of functions as returned by webpack blocks.
 * @return {Function}
 */
function when(condition, configSetters) {
  //assertConfigSetters(configSetters)
	if (condition) {
		return group(configSetters)
	} else {
		return () => (config) => config
	}
}
```

This could also allow reimplementing `env` like this (although it's already so simple this probable doesn't help at all)

```js
function env(envName, configSetters) {
     const currentEnv = process.env.NODE_ENV || 'development'
     return when(envName === currentEnv, configSetters);
}
```

andywer:
Hey @boxfoot! Thanks for sharing.

Having a `when` block seems to provide a nicer syntax than using the ternary operator, `.concat()` and friends. I am just not sure if adding such a thing to the core is a terrible idea or actually pretty handy...

@sapegin Any thoughts on this?

sapegin:
I like it ;-)

vlad-zhukov:
Fixed in #243

PR #243, opened 2017-12-07
Add when() block

boxfoot:
Adds a `when()` block to accept any conditional, as proposed in #242 .

I haven't reimplemented any other blocks (like `env()`) to use this one.

### DIFF
--- a/packages/core/lib/when.js
+++ b/packages/core/lib/when.js
@@ -1,0 +1,21 @@
+const { assertConfigSetters } = require('./configSetters')
+const group = require('./group')
+
+module.exports = when
+
+/**
+ * Applies an array of webpack blocks only if `condition` is true.
+ *
+ * @param {boolean} condition         Condition to test
+ * @param {Function[]} configSetters  Array of functions as returned by webpack blocks.
+ * @return {Function}
+ */
+function when (condition, configSetters) {
+  assertConfigSetters(configSetters)
+
+  if (condition) {
+    return group(configSetters)
+  } else {
+    return () => config => config
+  }
+}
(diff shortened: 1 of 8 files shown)


sapegin:
Cool, thanks!

vlad-zhukov:
There should be a note in the changelog about it.

packages/core/lib/when.js
```diff
@@ -0,0 +1,21 @@
...
+ * Applies an array of webpack blocks only if `condition` is true.
+ *
+ * @param {boolean} condition         Condition to test
+ * @param {Function[]} configSetters  Array of functions as returned by webpack blocks.
+ * @return {Function}
+ */
+function when (condition, configSetters) {
+  assertConfigSetters(configSetters)
+
+  if (condition) {
```
vlad-zhukov: Would be nice to have a check above that `condition` is a boolean.

boxfoot: I had intentionally *not* done this, on the logic that `condition` could be any truthy or falsy value...

andywer: I'm fine with `if (condition)`

vlad-zhukov: I am for consistency between code and docs. If we allow any values and cast them to booleans then docs should say that.

andywer:
Applied a minimal change to the README to reflect that truthy conditions will also work. Everyone happy now? :)

vlad-zhukov:
The block wasn't exported in `core/lib/index.js`.

andywer:
@boxfoot Can you please make sure that the `when()` block is actually exported (see `index.js`)?

andywer:
Should be fine to merge now.

boxfoot:
Thanks, @andywer  -- sorry for the slow response, I've just been swamped!

andywer:
No worries, I know the feeling 😉

Issue #246, opened 2018-01-06
Add tsd


zaaack:
Please add typescript definitions.

andywer:
Fair enough. Feel free to open a PR if it‘s an urgent issue 😉

andywer:
We need to discuss terminology here, since most terms haven't ever been clearly defined.
Current state of affairs is:

> Initially I was thinking that a `block` is a self-contained, configurable portion of a Webpack Configuration - so actually `entryPoint` is the block, not `entryPoint()`, which could be a `BlockInstance` or similar.
> 
> Then a `block` is everything that has the signature `(...args: any[]): BlockInstance`, where `BlockInstance = (current) WebpackBlock`

If someone has a better naming suggestion for `BlockInstance`, feel free to share it here 😉

andywer:
Another term that might be discussable is `WebpackBlockUpdater` (see PR #247).

Maybe it's rather a `WebpackConfigUpdater`: It's actually so generic that it's not webpack-blocks-specific. And this name might transport more information. Objections?

zcei:
`WebpackConfigUpdater ` is way better indeed. As it's just `(previousConfig: WebpackConfig): WebpackConfig` it really just is an updater for *any* Webpack config

andywer:
Suggestion for terminology, based on @zcei's PR:
- `entryPoint`, for instance, is the `Block`
- `entryPoint()` could be called a `BlockInstance`
- `WebpackConfig` should be self-describing
- `WebpackConfigUpdater` describes a function `(prevConfig: WebpackConfig) => WebpackConfig`

Note that I called it `Block`, not `WebpackBlock`, but yet called it `WebpackConfig` (not `Config`).

Idea behind it: There is no other "block" concept here than the lib's webpack blocks, so there might not be a need to make the name long. "config" on the other hand is such a general term that it might be better to call it "webpack config" when we talk about it.

Feel free to disagree now 👇

zcei:
I just pushed some changes to the PR (#247), where I went with my proposal from last post.

In fact we never specify `entryPoint` types, they all just inherently comply to the `Block` / `BlockCreator` interface, as they all return an instance.

So the term `Block` belongs to the instance (`entryPoint()`) for me, as it will way more often be used than the creator interface.

Everyone who'll be talking in issues about "a bug in the block" will talk about a specific configuration, so basically there's a bug in the instance of the created block.

As @andywer said:
> Feel free to disagree now 👇

PR #247, opened 2018-01-07
Add TypeScript typings for core & webpack

zcei:
Hej guys 👋 

I started tackling #246, for now only `@webpack-blocks/core` & `@webpack-blocks/webpack` to have a foundation, so we can discuss naming etc.

Afterwards we just need to add the proper `index.d.ts` to the blocks.

If I'm not mistaken there shouldn't be other changes necessary, at least VSCode gave me type hints in the sample app already, but I'll list some here for research:

* [ ] add `typings` to `package.json`(should be obsolete, as we comply to `index.d.ts` convention)
* [ ] add `index.d.ts` to `webpack-blocks` which exports types from sub-modules (should be infered from sub-module typings)


andywer:
Thanks for starting work on this, @zcei! Will have a deeper look into it :)

packages/webpack/index.d.ts
```diff
@@ -0,0 +1,27 @@
...
+  env,
+  match,
+  WebpackBlock,
+  WebpackBlockUpdater
+} from '@webpack-blocks/core'
+import { Configuration, Plugin, Entry, Options, Resolve, Output } from 'webpack'
+export { createConfig, group, env, match }
+
+export function addPlugins(plugins: Plugin[]): WebpackBlock
+export function customConfig(configSnippet: Configuration): WebpackBlockUpdater
```
zcei: Actually returns a `WebpackBlock`

andywer: I think we have to agree upon some terminology. The term `block`, for instance, has always been a bit fuzzy. Providing explicit types is probably the ideal time to define it properly.

zcei: Initially I was thinking that a `block` is a self-contained, configurable portion of a Webpack Configuration - so actually `entryPoint` is the block, not `entryPoint()`, which could be a `BlockInstance` or similar.

Then a `block` is everything that has the signature `(...args: any[]): BlockInstance`, where `BlockInstance = (current) WebpackBlock`

andywer: Fair enough. I would follow your argumentation here, @zcei. `BlockInstance` is not very catchy, but should do for now.

andywer:
Update: Let's better discuss this in the issue (#246) than in the PR.

PR #249, opened 2018-01-26
Enable Travis caching

vlad-zhukov:
This should speedup builds.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,3 +2,6 @@ language: node_js
 node_js:
   - node    # latest
   - 6
+cache:
+  directories:
+    - $HOME/.npm


andywer:
Hmm, doesn't seem to speed things up (just rebuilt to see if the 2nd build with enabled cache is faster now). Could still merge it, but I wonder why it's not faster.

vlad-zhukov:
No idea 😒 
Can we merge it and see how it goes? Does npm ever use caches? 😅

andywer:
We could also try yarn... But I am ok merging it, since I don't see any downsides.

vlad-zhukov:
Travis suggests to use `node_modules` for npm caching, but [it looks like it has its downsides](https://stackoverflow.com/questions/42521884/should-i-have-travis-cache-node-modules-or-home-npm). That's why I used `$HOME/.npm` here.

PR #250, opened 2018-01-29
Change match() to pass all webpack rule options to context.

vlad-zhukov:
Currently it's not possible to pass [webpack rule options](https://webpack.js.org/configuration/module/) other than `exclude` and `include` to blocks. Also there is no way to remove default options from blocks (for example the `exclude` in the `babel` block).

This PR fixes both problems.

### DIFF
--- a/packages/core/CHANGELOG.md
+++ b/packages/core/CHANGELOG.md
@@ -1,5 +1,9 @@
 # @webpack-blocks/core - Changelog
 
+## master
+
+- Change `match()` to passes all [webpack rule options](https://webpack.js.org/configuration/module/) to context. ([#250](https://github.com/andywer/webpack-blocks/pull/250))
+
 ## 1.0.0-beta.2
 
 - More useful error message when passing invalid blocks to `createConfig()` ([#171](https://github.com/andywer/webpack-blocks/issues/171))
--- a/packages/core/CHANGELOG.md
+++ b/packages/core/CHANGELOG.md
@@ -1,8 +1,8 @@
 # @webpack-blocks/core - Changelog
 
-## master
+## Next version
 
-- Change `match()` to passes all [webpack rule options](https://webpack.js.org/configuration/module/) to context. ([#250](https://github.com/andywer/webpack-blocks/pull/250))
+- Change `match()` to pass all [webpack rule options](https://webpack.js.org/configuration/module/) to context. ([#250](https://github.com/andywer/webpack-blocks/pull/250))
 
 ## 1.0.0-beta.2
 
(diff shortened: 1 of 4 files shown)


packages/core/CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # @webpack-blocks/core - Changelog
 
+## master
+
+- Change `match()` to passes all [webpack rule options](https://webpack.js.org/configuration/module/) to context. ([#250](https://github.com/andywer/webpack-blocks/pull/250))
```
vlad-zhukov: passes -> pass

vlad-zhukov:
> Also there is no way to remove default options from blocks (for example the exclude in the `babel` block).

I was wrong, the `exclude` option can be removed with `[]`. Anyway, I don't think it should be like that.

packages/core/lib/match.js
```diff
@@ -23,20 +21,13 @@ function match (test, options, configSetters) {
 
   assertConfigSetters(configSetters)
 
-  const match = { test: createFileTypeMatcher(test) }
+  options.test = createFileTypeMatcher(test)
```
vlad-zhukov: We should print a warning if `test` was set in options.

packages/core/lib/match.js
```diff
@@ -10,8 +10,6 @@ module.exports = match
  *
  * @param {string|RegExp|Function|Array} test   A glob like `*.css` or `{*.js, *.jsx}` or something else to use as `loader.test`.
  * @param {object} [options]                    Optional advanced matching options.
- * @param {string|Function|RegExp|Array|object} [options.include]
```
andywer: If we remove those hints, then we definitely need to improve the documentation of the overall `options` parameter, since otherwise it's not clear anymore what kind of options can be passed here 😉

packages/core/lib/match.js
andywer: Yeah, if `options.test` was set before this line 👍

packages/core/CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # @webpack-blocks/core - Changelog
 
+## master
```
vlad-zhukov: master -> Next version (like we used before)

vlad-zhukov:
Updated

packages/core/lib/match.js
```diff
@@ -21,6 +21,10 @@ function match (test, options, configSetters) {
 
   assertConfigSetters(configSetters)
 
+  if (options.test) {
+    console.warn(`match(): Setting 'test' in options is not supported and will be overriden with a 'test' argument.`)
```
andywer: Maybe we should even throw! What do you think?

Seems useful to defuse the bomb instead of putting a warning sign on it.

PR #253, opened 2018-02-15
Support negation globs in match()

### LINKED ISSUE #252: Allow negations in match
Taken from the readme:
```js
match('*.css', { exclude: path.resolve('node_modules') }, [
  css(),
]),
```
I think this would be better rewritten as 
```js
match(['*.css', '!*node_modules*'], [
  css(),
]),
```
or, in my use-case 
```js
match(['*.css', '!*.module.css'], [
  css(),
]),
```

Looking at the code, I think the negations could be extracted from the array and passed to the exclude here:

https://github.com/andywer/webpack-blocks/blob/d84ddd20b35ed8fc68f7ebd3e9e565e49d95db39/packages/core/lib/match.js#L29

Webpack accepts regexes fine in the exclude field.

What do you guys think? I can take a stab at it.

marcofugaro:
Closes #252

### DIFF
--- a/packages/core/lib/match.js
+++ b/packages/core/lib/match.js
@@ -4,6 +4,11 @@ const { invokePreHooks, invokePostHooks } = require('./hooks')
 
 module.exports = match
 
+const concat = (...arrays) => arrays.reduce((concatenation, array) => concatenation.concat(array), [])
+const regexify = glob => globToRegex(glob, { extended: true })
+const stripArrayConditionally = array => array.length === 1 ? array[0] : array
+const toArray = thing => Array.isArray(thing) ? thing : [ thing ]
+
 /**
  * State on which files to apply the loader blocks passed in this call. Works
  * like `group()`, but adds the file matching information to the context.
@@ -15,7 +20,7 @@ module.exports = match
  * @param {Function[]} configSetters            Array of functions as returned by webpack blocks.
  * @return {Function}
  */
-function match (patterns, options, configSetters) {
+function match (test, options, configSetters) {
   if (!configSetters && Array.isArray(options)) {
     configSetters = options
     options = {}
@@ -23,16 +28,17 @@ function match (patterns, options, configSetters) {
 
   assertConfigSetters(configSetters)
 
-  const match = {}
+  const { inclusions, exclusions } = splitPatterns(toArray(test))
+  const match = Object.assign({}, options, {
+    // The `.test` is usually just one pattern, so stripping the array feels more natural
+    test: stripArrayConditionally(normalizeMatchers(inclusions))
+  })
 
-  const [test, negations] = splitPatterns(patterns)
-
-  match.test = getMatches(patterns, test)
-  if (negations || options.exclude) {
-    match.exclude = getExcludes(patterns, negations, options.exclude)
-  }
-  if (options.include) {
-    match.include = options.include
+  if (exclusions.length > 0) {
+    match.exclude = concat(
+      match.exclude ? toArray(match.exclude) : [],
+      normalizeMatchers(exclusions)
+    )
   }
 
   const groupBlock = context => config => invokeConfigSetters(configSetters, deriveContextWithMatch(context, match), config)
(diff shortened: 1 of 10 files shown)


packages/core/lib/match.js
```diff
@@ -15,18 +15,21 @@ module.exports = match
...
   }
 
   assertConfigSetters(configSetters)
 
-  const match = { test: createFileTypeMatcher(test) }
+  const match = {}
 
-  if (options.exclude) {
-    match.exclude = options.exclude
+  const [test, negations] = splitPatterns(patterns)
```
andywer: I would rather put the `typeof patterns === 'string'` here. Would make the splitPatterns() function much simpler.

And I think there are basically just two use cases:
1. `patterns: string[]`
2. `patterns: string`

andywer: Maybe we should also still call it `test`, since `patterns` sounds like `string[]`.

👉 `splitPatterns(typeof test === 'string' ? [test] : test)`

packages/core/lib/match.js
```diff
@@ -42,16 +45,56 @@ function match (test, options, configSetters) {
 
 const regexify = glob => globToRegex(glob, { extended: true })
 
-function createFileTypeMatcher (test) {
-  if (typeof test === 'string') {
+// split positive matches and negative matches
+function splitPatterns (patterns) {
```
andywer: Why doesn't it just return `[ inclusions, exclusions ]`? 😉

packages/core/lib/match.js
```diff
@@ -15,18 +15,21 @@ module.exports = match
...
-  const match = { test: createFileTypeMatcher(test) }
+  const match = {}
 
-  if (options.exclude) {
-    match.exclude = options.exclude
+  const [test, negations] = splitPatterns(patterns)
+
+  match.test = getMatches(patterns, test)
+  if (negations || options.exclude) {
+    match.exclude = getExcludes(patterns, negations, options.exclude)
```
andywer: This is all quite complicated. Could be simplified:

```js
const [ inclusions, exclusions ] = splitPatterns(/* ... */)
match.exclude = makeArray(options.exclude).concat(exclusions)
```

packages/core/lib/match.js
andywer: Or `{ inclusions, exclusions }`.

andywer:
Thanks for your PR, Marco. Maybe I can find some time later today to code a bit if you want.

packages/core/lib/match.js
marcofugaro: `{ inclusions, exclusions }` looks better!

marcofugaro:
Sure! Some help would be nice!

andywer:
@marcofugaro Won't make it today. Will help tomorrow 😉

andywer:
Done. Think it's way simpler now :)

Would also love to see a quick review from another senior contributor ✌️
@vlad-zhukov @sapegin

andywer:
That was fast! Thx, @jvanbruegge 😊

Leaving the PR open for a little longer, so everyone has a chance to chip in.

packages/core/lib/match.js
```diff
@@ -4,6 +4,11 @@ const { invokePreHooks, invokePostHooks } = require('./hooks')
 
 module.exports = match
 
+const concat = (...arrays) => arrays.reduce((concatenation, array) => concatenation.concat(array), [])
```
vlad-zhukov: nit: can use `_.concat` instead.

vlad-zhukov:
Looks good!

packages/core/lib/match.js
```diff
@@ -23,13 +28,17 @@ function match (test, options, configSetters) {
...
+  })
 
-  if (options.exclude) {
-    match.exclude = options.exclude
-  }
-  if (options.include) {
-    match.include = options.include
+  if (exclusions.length > 0) {
+    match.exclude = _.concat(
+      match.exclude ? toArray(match.exclude) : [],
```
vlad-zhukov: That's going to filter ligitimate values like `null`.

vlad-zhukov: Nevermind, `exclude` will always be an array anyway.

marcofugaro:
@andywer I updated all the docs containing `match`

PR #254, opened 2018-02-20
Support for ESLint

dmitmel:
Almost like the `tslint` block, but for JavaScript.

### DIFF
--- a/packages/eslint/CHANGELOG.md
+++ b/packages/eslint/CHANGELOG.md
@@ -1,5 +1,5 @@
 # @webpack-blocks/eslint - Changelog
 
-## 1.0.0-rc
+## Next release
 
 Initial release.
--- a/packages/eslint/CHANGELOG.md
+++ b/packages/eslint/CHANGELOG.md
@@ -1,0 +1,5 @@
+# @webpack-blocks/eslint - Changelog
+
+## 1.0.0-rc
+
+Initial release.
--- a/packages/eslint/CHANGELOG.md
+++ b/packages/eslint/CHANGELOG.md
@@ -1,5 +1,5 @@
 # @webpack-blocks/eslint - Changelog
 
-## Next release
+## 1.0.0-rc
 
 Initial release.
--- a/packages/eslint/README.md
+++ b/packages/eslint/README.md
@@ -1,4 +1,4 @@
-# webpack-blocks - TSLint
+# webpack-blocks - ESLint
 
 [![Gitter chat](https://badges.gitter.im/webpack-blocks.svg)](https://gitter.im/webpack-blocks)
 [![JavaScript Style Guide](https://img.shields.io/badge/code%20style-standard-brightgreen.svg)](http://standardjs.com/)
--- a/packages/eslint/README.md
+++ b/packages/eslint/README.md
@@ -1,0 +1,35 @@
+# webpack-blocks - TSLint
+
+[![Gitter chat](https://badges.gitter.im/webpack-blocks.svg)](https://gitter.im/webpack-blocks)
+[![JavaScript Style Guide](https://img.shields.io/badge/code%20style-standard-brightgreen.svg)](http://standardjs.com/)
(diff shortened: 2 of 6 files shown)


packages/eslint/README.md
```diff
@@ -0,0 +1,35 @@
+# webpack-blocks - TSLint
```
zcei: Copy 🍝 - needs to be `ESLint` as well 🙂

dmitmel: Oh, thanks :) These packages are very similar, so I've decided to copy files from the `tslint` block.

zcei: Sure, that's how I would have done as well 😉

andywer:
Proposal: Let's extract the `tslint` block out of the core webpack-blocks repo into its own and let's put the `eslint` block into a separate repo as well.

We can still link to them in the README here, but trimming down the number of packages here might be a good idea. Thoughts?

@jvanbruegge

dmitmel:
Well, I think these packages must be included in the `webpack-blocks` package because they are used in lots of projects (see [their](https://npm.im/eslint-loader) [stats](https://npm.im/tslint-loader)). But, [your CI can't run all tests](https://travis-ci.org/andywer/webpack-blocks/jobs/344207521#L1132-L1135) because there are too many packages. An ideal solution would be creating an organization and putting all packages into separate repositories.

jvanbruegge:
@andywer
Would be ok with that, just dont forget to make a major version bump on the `webpack-blocks` package. I dont want all the setups suddenly breaking

zcei:
I would really like to see most blocks in their own repo - that's why I initially blocked the [`webpack-blocks` org](https://github.com/webpack-blocks) org.
Though I think especially ESLint (& TSLint ?) should be treated first-class as they're nearly always used.

andywer:
Then maybe let's postpone this decision (extra repo or not) for now 😉

@jvanbruegge That shouldn't be a breaking change. Should have no influence on user's where the sources reside.

@dmitmel Yeah, but the block's are more utility than core functionality and the [download count of the tslint block](https://www.npmjs.com/package/@webpack-blocks/tslint) is quite low.

The CI issue is quite annoying, though... We could try `lerna run --parallel` or `lerna run --stream` to circumvent that issue or try a different monorepo tool.

dmitmel:
@andywer Would you like to add the eslint block or remove the tslint block?

andywer:
@dmitmel Let's add the eslint block for now and decide on the overall strategy afterwards. That discussion shouldn't be a reason to block new features :)

packages/eslint/CHANGELOG.md
```diff
@@ -0,0 +1,5 @@
+# @webpack-blocks/eslint - Changelog
+
+## Next release
```
andywer: In this case we can enter the version number already, since we will publish right away.

vlad-zhukov:
I'd move unpopular and framework-specific blocks out of the main repo (for example `elm`). It will make installs and tests faster (good for both development and CI), but most importantly you can grant write access to people who care about the block, and don't block their development interests.

vlad-zhukov:
@dmitmel Thanks for the PR! Would be great to have a few test cases similar to ones other blocks have.

dmitmel:
@vlad-zhukov I've written some [integration tests](https://github.com/dmitmel/webpack-blocks/blob/feature/eslint-block/packages/eslint/__tests__/eslint.test.js). Thanks for pointing out!

andywer:
Awesome, thanks for your efforts! 😊

andywer:
Published as `@webpack-blocks/eslint@1.0.0-rc` 🚀

dmitmel:
[I've found something wrong](https://github.com/dmitmel/webpack-blocks/commit/71f0940c95d81d43ed11a5fcff3a3f6a739e5068) in the eslint block: it has `@webpack-blocks/core` in its peer dependencies.

andywer:
Might be a copy-pasta mistake. I don't remember with certainty, but I think at some point we added that peer dependency to make explicit that the block only works with webpack-blocks v1 (not v0.x).

It's not critical and the peer dependency is basically true 😉

dmitmel:
@andywer Nice ~hack~ workaround!

Issue #262, opened 2018-03-27
Code style


dmitmel:
While refactoring code, I noticed that the codebase contains _different coding styles_ because many people contribute to this project. This causes some problems like inconsistent line length.

My suggestions are:

- **Use [Prettier](https://prettier.io/) instead of Standard**, because Prettier produces more _consistent_ code. However, Prettier doesn't follow the JavaScript Standard Style, e.g. it doesn't put a space after function name in declarations.
- **Use ESLint** with something like [the Airbnb preset](https://github.com/airbnb/javascript).

andywer:
Could you please link to those places with different code styles?

Using prettier has been a topic for quite some time. I don't see any reason against, there was just no time yet to do that.

How would the Airbnb ruleset change anything? Because it's stricter?

dmitmel:
@andywer

> Using prettier has been a topic for quite some time. I don't see any reason against, there was just no time yet to do that.

Setting up Prettier and ESLint is really easy, I've even created an [ESLint preset](https://github.com/dmitmel/eslint-preset-dmitmel) that automatically configures Prettier and the Airbnb preset.

> How would the Airbnb ruleset change anything? Because it's stricter?

No. It's just my preference.

farism:
I prefer husky + prettier and no eslint. I understand that eslint handles a whole class of things that prettier does not, but I kind of like to talk about those things during code review anyway.

andywer:
> I understand that eslint handles a whole class of things that prettier does not, but I kind of like to talk about those things during code review anyway.

That is true for code style, but the important "whole class of things that prettier does not handle" is stuff like checking for unused variables, broken references, etc. That's what prettier cannot provide. I say it should not be the reviewer's responsibility either, since it can be reliably checked by software 😉

andywer:
Is this issue still relevant, now that we have prettier in the v2.0 branch?

vlad-zhukov:
We've got both Prettier and ESLint. Now we can add more ESLint rules over time.

PR #263, opened 2018-04-01
Replace standard with prettier

jvanbruegge:
Related to #262 

with this the formatter runs on `npm test` (in the root) and on any modified file (via pre-commit hook)

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,12 +1,26 @@
 {
+  "husky": {
+    "hooks": {
+      "pre-commit": "lint-staged"
+    }
+  },
+  "lint-staged": {
+    "*.{js,ts,tsx}": [
+      "prettier-standard",
+      "git add"
+    ]
+  },
   "scripts": {
-    "test": "lerna run test",
+    "test": "npm run format && lerna run test",
+    "format": "prettier-standard 'packages/**/*.{js,ts,tsx}'",
     "postinstall": "lerna bootstrap && link-parent-bin"
   },
   "devDependencies": {
     "ava": "^0.22.0",
+    "husky": "^0.14.3",
     "lerna": "2.0.0-beta.30",
     "link-parent-bin": "^0.1.3",
-    "standard": "^10.0.3"
+    "lint-staged": "^7.0.2",
+    "prettier-standard": "^8.0.0"
   }
 }
--- a/package-lock.json
+++ b/package-lock.json
@@ -69,6 +69,303 @@
         "slide": "1.1.6"
       }
     },
+    "@babel/code-frame": {
+      "version": "7.0.0-beta.42",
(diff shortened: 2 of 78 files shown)


package.json
```diff
@@ -1,12 +1,25 @@
 {
+  "husky": {
+    "hooks": {
+      "pre-commit": "lint-staged"
+    }
+  },
+  "lint-staged": {
```
dmitmel: You forgot to add `lint-staged` to dev dependencies

package.json
```diff
@@ -1,12 +1,25 @@
...
+    "*.{js,ts}": [
```
dmitmel: There is a file called `packages/sample-app/src/App.tsx`, so the `.tsx` extension should also be added to this list

package.json
```diff
@@ -1,12 +1,25 @@
...
+      "prettier --no-semi --single-quote --write",
+      "git add"
+    ]
+  },
   "scripts": {
-    "test": "lerna run test",
+    "test": "npm run format && lerna run test",
+    "format": "prettier --no-semi --single-quote --write 'packages/**/*.{js,ts}'",
```
dmitmel: Same with extensions as [here](https://github.com/andywer/webpack-blocks/pull/263/files#diff-b9cfc7f2cdf78a7f4b91a753d10865a2R8)

package.json
dmitmel: Ok, this is done

package.json
```diff
@@ -1,12 +1,26 @@
 {
+  "husky": {
+    "hooks": {
+      "pre-commit": "lint-staged"
+    }
+  },
+  "lint-staged": {
+    "*.{js,ts,tsx}": [
+      "prettier --no-semi --single-quote --write",
```
dmitmel: You can configure Prettier via `prettier` key in the `package.json`, so let's move prettier config there to reduce duplication (see [line 15](https://github.com/andywer/webpack-blocks/pull/263/commits/f2dc22756be23685e767f20196e79aaeb8fb9c26#diff-b9cfc7f2cdf78a7f4b91a753d10865a2R15))

dmitmel:
Should we format Markdown files with Prettier?

andywer:
Might also make sense to use [`prettier-standard`](https://github.com/sheerun/prettier-standard)?

dmitmel:
@andywer ESLint has more linting options than Standard, but for now we can use Standard.

andywer:
@dmitmel Of course, but I wouldn't change everything at once. Incremental changes are way easier to handle 😉

jvanbruegge:
Changed it to use `prettier-standard`

jvanbruegge:
It looks like prettier-standard does not fix all its issues, which kinda defeats the point of prettier

vlad-zhukov:
Superseded by #275.

PR #268, opened 2018-04-21
Use yarn workspaces

dmitmel:


.gitignore
```diff
@@ -1,4 +1,5 @@
 node_modules/
 packages/*/node_modules/
-lerna-debug.log
-npm-debug.log
+lerna-*.log
+npm-*.log
+yarn-*.log
```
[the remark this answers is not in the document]
dmitmel: > Could you please change these 3 to simply `*.log`?

Sure.

> And also would be nice to add `package-lock.json`.

Could you please explain why?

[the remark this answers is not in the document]
vlad-zhukov: To make sure it won't be accidentally added in the future when someone installs deps with `npm`.

package.json
```diff
@@ -1,12 +1,16 @@
 {
+  "private": true,
   "scripts": {
-    "test": "lerna run test",
-    "postinstall": "lerna bootstrap && link-parent-bin"
```
andywer: Why not hook up on the `postinstall` hook?

Btw, we might not need `lerna bootstrap` anymore if we run `yarn` and have workspaces. We still need `link-parent-bin`, though, since yarn also tends to fail to link bins installed in top-level node_modules.

vlad-zhukov: Are you sure yarn fails to link bins? Is there a way to test it?

andywer: I set-up a few monorepos with yarn workspaces in the last couple of month and encountered that issue at least twice.

Not sure if there is a real way to test. But if a package has a dependency with a CLI tool (`ava` for instance) and the dependency is installed in the monorepo-wide `node_modules`, then I think the tool will always be missing from the package`s `node_modules/.bin` as of now.

andywer: But the tests are green... 🤔

packages/webpack/__e2e-fixtures__/typescript/app.ts
```diff
@@ -1,4 +1 @@
-declare var module : any
-declare var process : any
-
 module.exports = process.env.TEST
```
zcei: I'd like to keep having something TypeScript specific in the `app.ts` - currently we set the webpack config to read from `app.ts` and have the ts block set up, but I think it would also work without, as it currently is valid JS.

Probably a simple 
```ts
module.exports = process.env.TEST as string
```
would be enough

package.json
vlad-zhukov: I've just installed this branch and yarn didn't create links for packages that are not in the root, so I guess this is the default behaviour. Why do you need it for by the way?

packages/webpack/__e2e-fixtures__/typescript/app.ts
dmitmel: @zcei What about this?

```ts
function add(a: number, b: number): number {
  return a + b;
}
```

package.json
dmitmel: Yarn documentation says:

> `[script]` can also be any locally installed executable that is inside `node_modules/.bin/`.

packages/webpack/__e2e-fixtures__/typescript/app.ts
zcei: Even better

package.json
andywer: > I've removed the postinstall hook because it causes an infinite loop (lerna/lerna#1125).

Good catch! Question is if we even need `lerna bootstrap` anymore. What's the benefit of `lerna bootstrap` instead of just running `yarn`? It will also install the root dependencies, all package dependencies and link the packages.

dmitmel: > Question is if we even need `lerna bootstrap` anymore

So, should I remove the `bootstrap` script from the root package?

andywer: Let's give it a try :)

(I can do it, too, if you want; in the evening)

dmitmel: > Let's give it a try :)

Sorry, @andywer, what do you mean?

andywer: I meant
> So, should I remove the bootstrap script from the root package?

👉I think it should still work without `lerna bootstrap` when using yarn and yarn workspaces. Let's try and see if that's so 😉

andywer: Good enough for me 😅

But seriously, I think it does pretty much exactly the same, so let's get rid of one of them.

farism: `lerna bootstrap` actually just [uses yarn install](https://github.com/lerna/lerna/issues/1021#issuecomment-333598007) when using workspaces, so +1 to killing any references to it.

vlad-zhukov:
Updated and merged. Thanks everyone for working on it!

PR #270, opened 2018-04-23
Remove stale package-lock.json files

farism:
Now that `webpack-blocks` is using yarn workspaces we should remove any `package-lock.json` files that are generated by NPM. NPM technically is no longer supported by `webpack-blocks`.


zcei:
@vlad-zhukov I thought the same, but no - #268 only removed huge parts of the package-lock, but not the file completely. That was a bit missleading from GitHub's collapsed view.

vlad-zhukov:
I bet these are your local files. You can go through the [`release-2.0`](https://github.com/andywer/webpack-blocks/tree/release-2.0) branch on GitHub to make sure there are no `package-lock.json`s left!

vlad-zhukov:
I've just made a clean git pull and the only `package-lock.json` left is in the `eslint` package. Interestingly both this PR and #268 missed it.

vlad-zhukov:
Actually I was looking at a different branch, there are not `package-lock.json` files left.

PR #272, opened 2018-04-25
Update Travis config

vlad-zhukov:


.travis.yml
```diff
@@ -1,6 +1,17 @@
...
-script: travis_wait 15 yarn test   # Increase "no output received" timeout
+
+cache:
+  yarn: true
+  directories:
+    - ".eslintcache"
+    - "node_modules"
+
+before_install:
+  - curl -o- -L https://yarnpkg.com/install.sh | bash
```
zcei: This should only be needed for a specific version, no? Or Is Travis lacking behind in the Yarn version it supports?

vlad-zhukov: Travis has 1.3.2, this script updates it to the latest.

PR #276, opened 2018-05-08
Hoist ava to the root

vlad-zhukov:
And also update it to the latest.


package.json
```diff
@@ -3,11 +3,11 @@
   "scripts": {
     "lint": "eslint \"**/*.js\"",
     "format": "prettier \"**/*.{js,ts,tsx}\" --write --loglevel warn",
-    "test": "prettier \"**/*.{js,ts,tsx}\" --list-different && yarn lint && lerna run test"
+    "test": "prettier \"**/*.{js,ts,tsx}\" --list-different && yarn lint && ava"
```
andywer: I know, it's not exactly the scope of this PR, but:

How about moving `yarn lint`, for instance, into a `posttest` script? Might make the scripts easier readable. Thoughts?

packages/sample-app/__tests__/__snapshots__/webpack-config.test.js.md
```diff
@@ -9,33 +9,16 @@ Generated by [AVA](https://ava.li).
 > Snapshot 1
 
     {
-      devServer: {
```
andywer: Is that correct? ... 🤔

packages/sample-app/__tests__/webpack-config.test.js
```diff
@@ -8,7 +8,23 @@ const configFile = require.resolve('../webpack.config.babel')
 
 test('it exports the development config', t => {
   const originalConfig = requireConfig()
+  const module = Object.assign({}, originalConfig.module, {
```
andywer: Didn't really expect that here. It's surely not wrong, but it there a background story why you added this just now? (Just curious :) )

packages/sample-app/__tests__/__snapshots__/webpack-config.test.js.md
vlad-zhukov: It changed when I only updated `ava` from 0.22.0 to 0.25.0 before doing any other changes. I couldn't find any [release notes](https://github.com/avajs/ava/releases) related to it.

I'll have look on how to fix it.

package.json
vlad-zhukov: {}

vlad-zhukov:
```

Comment (vlad-zhukov) [reply package.json]:
```

vlad-zhukov:
```
{
  "test": "prettier \"**/*.{js,ts,tsx}\" --list-different && yarn lint && ava"
}
```
vs
```
{
  "pretest": "prettier \"**/*.{js,ts,tsx}\" --list-different && yarn lint",
  "test": "ava"
}
```

It barely improves anything 😄 If you want to make it shorter I can replace `--list-different` with `-l`.

andywer: Ahhh, keep it 😅

packages/sample-app/__tests__/webpack-config.test.js
vlad-zhukov: Because I made a change in `packages/sample-app/webpack.config.babel.js`:

```diff
-  typescript(),
+  typescript({ configFileName: path.resolve(__dirname, './tsconfig.json') }),
```

packages/sample-app/__tests__/__snapshots__/webpack-config.test.js.md
vlad-zhukov: Fixed by setting `NODE_ENV` to `development`. This was broken because [since 0.23.0 `ava` sets `NODE_ENV` to `test`](https://github.com/avajs/ava/releases/tag/v0.23.0).

packages/assets/package.json
```diff
@@ -5,10 +5,6 @@
   "main": "lib/index",
   "license": "MIT",
   "author": "Andy Wermke <andy@dev.next-step-software.com>",
-  "scripts": {
-    "test": "ava",
-    "prepublishOnly": "yarn test"
```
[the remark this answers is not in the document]
vlad-zhukov: Nope, manual only for now. I would like to tickle it in another PR.

- Publishing single packages leads to version inconsistencies and bugs. For instance the `uglify` block was publish separately as 1.0.0 and later as 1.1.0, when other blocks used to be v1.0.0-beta. Then with the 1.0 release the `uglify` was republished and its version 1.0.0 was tagged as latest. What that means is if you run `yarn add @webpack-blocks/uglify` now, you will get 1.0.0 instead of 1.1.0. (I am sorry I haven't reported it before 😅)
- Some blocks have no unit but instead have related e2e tests. I believe these are more important to have and we can focus on writing more of them.

packages/sample-app/__tests__/__snapshots__/webpack-config.test.js.md
```diff
@@ -1,4 +1,4 @@
-# Snapshot report for `__tests__/webpack-config.test.js`
+# Snapshot report for `packages\sample-app\__tests__\webpack-config.test.js`
```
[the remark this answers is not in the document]
vlad-zhukov: Nice catch! I've just regenerated snapshots on my Mac.

packages/assets/package.json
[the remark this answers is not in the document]
zcei: Oh okay. Then we should just create a small doc to refer to when publishing. Like that we only publish "whole monorepo", not standalone.

PR #259, opened 2018-03-23
API for version 2.0

dmitmel:
This PR shows how the new block API will look like.

Related to #125

PR #277, opened 2018-05-19
The new block API

vlad-zhukov:
I've proposed this API almost a year ago in #125 and I believe everyone is fine with it by now. TLDR:

```js
// Current:

function postConfig (context, util) {
  const ruleConfig = Object.assign({
    test: /\.(js|jsx)$/,
    exclude: /node_modules/,
    use: [
      { loader: 'babel-loader', options: context.babel }
    ]
  }, context.match)

  return util.addLoader(ruleConfig)
}

// This PR:

function postConfig (context) {
  const ruleConfig = Object.assign({
    test: /\.(js|jsx)$/,
    exclude: /node_modules/,
    use: [
      { loader: 'babel-loader', options: context.babel }
    ]
  }, context.match)

  return context.addLoader(ruleConfig)
}
```

The #259 is different because it merges `context` and utils into an object:

```js
function postConfig ({context, addLoader}) {
  // ...
}
```

cc @andywer @zcei @dmitmel @jvanbruegge

### DIFF
--- a/packages/assets/lib/css.js
+++ b/packages/assets/lib/css.js
@@ -17,8 +17,8 @@ function css(options = {}) {
     loaders.unshift({ loader: 'style-loader', options: options.styleLoader || {} })
   }
 
-  return (context, util) =>
-    util.addLoader(
+  return context =>
+    context.addLoader(
       Object.assign(
         {
           test: /\.css$/,
@@ -56,8 +56,8 @@ function cssModules(options = {}) {
     loaders.unshift({ loader: 'style-loader', options: options.styleLoader || {} })
   }
 
-  return (context, util) =>
-    util.addLoader(
+  return context =>
+    context.addLoader(
       Object.assign(
         {
           test: /\.css$/,
--- a/packages/assets/lib/file.js
+++ b/packages/assets/lib/file.js
@@ -6,7 +6,7 @@ module.exports = file
  * @see https://github.com/webpack-contrib/file-loader
  */
 function file(options = {}) {
-  return (context, util) => {
+  return context => {
     if (!context.match) {
       throw new Error(
         `The file() block can only be used in combination with match(). ` +
@@ -14,7 +14,7 @@ function file(options = {}) {
       )
     }
 
-    return util.addLoader(
+    return context.addLoader(
       Object.assign(
         {
           use: [{ loader: 'file-loader', options }]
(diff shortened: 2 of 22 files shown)


jvanbruegge:
Not sure what this is solving actually. AFAICT 2.0 old blocks to be used with more or less no change at all, so not sure if this is worth a breaking change

zcei:
So #259 and this PR are two competing approaches?

Same problem as @jvanbruegge, the mentioned discussion is too long and I don't recall all the details. Would you mind a small explanation what the problems are with the current API & what the up & downsides of the two competing PRs are?

Otherwise we can't make a profound decision, and I think some visibility helps when we later come back here (as it's linked in the changelog)

vlad-zhukov:
Good point! However I'm unsure v1 blocks will work with v2 out of the box w/o any changes. Webpack 4 made breaking changes for loaders and in order to use old blocks with v2 a developer would need to update dependencies at the very least. Making another single line change shouldn't be a big of a problem, is it?

The reason for the change is purely stylistic. Some people (including me) would like to be it that way. There is also a rare case when a block doesn't use the context but needs a util function: `(_, util) => util.addLoader({})`.

There are no hard feelings about the change. How often do people develop new blocks anyway? 🙃

andywer:
I think there was even a third option: Don't pass the util stuff (`addLoader` & friends) into the block at all, but rather make the blocks `import` them from a separate package.

(Sorry for the late reply!)

vlad-zhukov:
@andywer Was it even considered as an option? 🙉
Right now I'm very close to leaving the API as it is for now. Maybe next year...

andywer:
@vlad-zhukov Not sure how serious the efforts in that direction where, but I remember that one, too. And I totally get how you are feeling right now; this API discussion is a hornet's nest... 🙈

At some point I just stopped caring about it this API issue too much. Bottom line: No matter what we do here, the average webpack-blocks user will neither notice nor care 😜

andywer:
@dmitmel No need to be snappish. I am just trying to recall the discussion at the time. And yes, I had my reservations as well! Nevertheless it doesn't hurt to gather the whole picture.

dmitmel:
@andywer Ok, but I'm just trying to do something because I wasn't contributing to `webpack-blocks` for a long time.

andywer:
@dmitmel Trust me, your efforts are highly appreciated! It's just a tricky change that requires quite a lot of communication. If you want, we can all have a chat tomorrow on gitter.im about the open issues and maybe we can find a less controversial issue that you can help with 🙂

But enough of that now. All non-PR-related talk moves to gitter now.

andywer:
I'm sorry, but I don't see this happening at the moment. Closing for now.

Feel free to bring it up again 😉

PR #279, opened 2018-05-25
Fix a skipped test

vlad-zhukov:
While working on it I found out [Extract Text Plugin](https://github.com/webpack-contrib/extract-text-webpack-plugin) is deprecated for webpack 4 and they suggest to use [mini-css-extract-plugin](https://github.com/webpack-contrib/mini-css-extract-plugin).


packages/sample-app/webpack.config.babel.js
```diff
@@ -35,32 +34,30 @@ const developmentConfig = () =>
...
   group([
-    extractText(),
-    addPlugins([
-      new webpack.LoaderOptionsPlugin({
-        minimize: true,
-        debug: false
-      })
-    ])
+    css.modules({ minimize: true }),
+    extractText('css/[name].[contenthash:hex:8].css')
```
vlad-zhukov: Had to change it in order to fix the [`contenthash` issue](https://github.com/webpack-contrib/extract-text-webpack-plugin/issues/763).

andywer:
Tests fail, because of prettier if I see that correctly. Any idea?

andywer:
Btw, very good point. We should switch to using the `mini-css-extract-plugin`... 👍

vlad-zhukov:
All errors have been fixed.

@andywer I forgot to run Prettier and ESLint before committing 😐

PR #281, opened 2018-05-29
Format markdown files with Prettier

vlad-zhukov:


README.md
```diff
@@ -100,18 +99,15 @@ module.exports = createConfig([
     }),
     sourceMaps()
   ]),
-  env('production', [
-    uglify(),
-    addPlugins([
-      new webpack.LoaderOptionsPlugin({ minimize: true })
-    ])
-  ])
+  env('production', [uglify(), addPlugins([new webpack.LoaderOptionsPlugin({ minimize: true })])])
```
vlad-zhukov: It makes JS code less readable in some cases 😞

vlad-zhukov: We can wrap such cases with `prettier-ignore`: https://prettier.io/docs/en/ignore.html#range-ignore

jvanbruegge:
Did not know prettier can do this, especially correctly formatting inline code, so this is pretty rad :+1:

Issue #264, opened 2018-04-21
2.0 roadmap


vlad-zhukov:
Hello! I'm interested to release a v2.0 as soon as possible and I've got an approval from other maintainers to work on it on our gitter! ([Join it if you haven't yet!](https://gitter.im/webpack-blocks/Lobby))

---

This is going to be a breaking, **webpack-4-only** release, that updates the block API and drops deprecated ones. We also aim to make lots internal changes to make further maintaining much easier and beginner friendlier!

All new changes should go into the `release-2.0` branch (#266), while `master` stays the same and represents the current stable release (v1.0).

## TODO

- [x] Migrate to `webpack 4` (tracking issue #255).
- [x] Consider updating to `babel@7`.
- [x] Remove the deprecated `fileType` API (PR #260).
- [x] Remove the deprecated `resolveAliases` block. (PR #267)
- [x] Use `yarn` workspaces (PR #268).
- [x] Decide to either upgrade `lerna` or use something else (see discussion in #236).
- [x] Remove `ava` from devDeps of every package and hoist it to the root. (PR #276).
- [x] Use `prettier` (PR ~~#263~~ #275).
- [x] Consider adding some `eslint` code rules (not style ones).
- [ ] Other minor changes left in [2.0.0 Milestone](https://github.com/andywer/webpack-blocks/milestone/3).
- [x] Cut a branch `1.0.0` from `master` before merging `release-2.0`.
- [x] List contributors!

## Removed from 2.0 release scope

- [ ] Update the block API (see discussion in #125, PR #259).
- [ ] Consider moving unpopular blocks to the [`webpack-blocks` org](https://github.com/webpack-blocks) (issue #269).
- [ ] Add TypeScript definitions (issue #246, PR #247).
- [ ] Write a migration guide.
- [ ] Inform 3rd-party block maintainers!

@andywer @jvanbruegge @zcei @dmitmel @marcofugaro

vlad-zhukov:
@dmitmel I really want this to happen because I use babel 7 myself since `alpha`s. But they keep making breaking changes in `beta`s and suggest to pin exact versions themselves, which is fine for an app but absolutely not fine for a library. Once they announce there are not going to be breaking changes anymore (probably in `rc`s), we can update the block. But for now it's going to wait for some time.

vlad-zhukov:
I've created a `release-2.0` branch, all new changed should target it instead of the `master`.

dmitmel:
[...]

PR #282, opened 2018-06-14
Change default devtool to cheap-module-eval-source-map

pravdomil:
Because they provide same level of detail as `cheap-module-source-map` but they are faster in rebuild.

see https://webpack.js.org/configuration/devtool


vlad-zhukov:
Nice catch, thanks! Could you please point it to the `release-2.0` branch instead of the `master`? (See #264 for reference.) And after that you'd also need to make sure tests pass.

dmitmel:
@pravdomil I have a local copy of webpack-blocks so I can make a PR towards your fork with the required changes to snapshots.

pravdomil:
@dmitmel great

pravdomil:
> All checks have passed

pravdomil:
I have deleted the original repo so I have to open new PR https://github.com/andywer/webpack-blocks/pull/285

PR #283, opened 2018-06-14
use object spread

pravdomil:
see diff

### DIFF
--- a/README.md
+++ b/README.md
@@ -295,16 +295,17 @@ Take the `babel` webpack block for instance:
  * @param {RegExp|Function|string}  [options.exclude]   Directories to exclude.
  * @return {Function}
  */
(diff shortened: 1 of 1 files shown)


dmitmel:
Object rest spread was added in [Node 8](https://node.green/#ES2018-features-object-rest-spread-properties-object-spread-properties), but the minimal version of Node which webpack-blocks supports is 6th.

pravdomil:
ok

pravdomil:
btw: node 6 is only maintenance mode right now

https://github.com/nodejs/Release#release-schedule

dmitmel:
Yes, but it's still [the most popular](https://nodesource.com/node-by-numbers) version of Node.

pravdomil:
didn't know that, thanks

PR #287, opened 2018-06-19
Minor fixes

andywer:
Could not run the tests locally, due to two small issues:

- Missing `babel-eslint` dev dependency in @wb/webpack package
- Linted `elm-stuff` directory


andywer:
Hmm, that doesn't look right... Starting over.

PR #289, opened 2018-06-19
Bugfix: Drop awesome-typescript-loader PathPlugin

andywer:
... that "is available only for webpack@2.1". (See [their readme](https://github.com/s-panferov/awesome-typescript-loader#advanced-path-resolution-in-typescript-20))

Couldn't get the end-to-end tests to run successfully on my local machine. It would always fail when using the `TsConfigPathsPlugin` and would work whenever I remove it.

Not sure if this will have a major impact on the users.

Update: It should be fine, I think, since you are now able to pass the path options directly to the loader. The `typescript()` block even does that already.


andywer:
@jvanbruegge Any thoughts on this?

dmitmel:
@andywer Could you please provide "Steps to reproduce", so that others can test if they can get the same results on their machines?

andywer:
Damn it. Doesn't happen when doing it in a freshly cloned repo...

It was straight forward:
```sh
$ yarn
$ yarn test
```

And then when running the tests, the typescript end-to-end test hang and just wouldn't finish ever.

When I ran it in a freshly cloned repo, it worked 🙃

Sorry for bothering, guys. Still seems a bit strange, but since it works with a clean repo, I will just close it for now. Might be some stale untracked files, like in the former case. Will clean up my local installation...

andywer:
Another update: Works for a freshly cloned repo cloned to `/tmp/webpack-blocks`, but fails when cloning to `~/workspace/webpack-blocks` 🤯

```sh
$ cd /tmp
$ git clone git@github.com:andywer/webpack-blocks.git -b release-2.0
$ cd webpack-blocks
$ yarn
$ yarn test
$ # works
```

```sh
$ cd ~/workspace
$ rm -rf webpack-blocks
$ git clone git@github.com:andywer/webpack-blocks.git -b release-2.0
$ cd webpack-blocks
$ yarn
$ yarn test
$ # hangs after 66 tests, fan spins up
```

vlad-zhukov:
Weird, I am not able to reproduce.

PR #290, opened 2018-06-25
Upgrade deps

vlad-zhukov:


packages/sample-app/package.json
```diff
@@ -10,22 +10,23 @@
     "node": ">= 6.0"
   },
   "devDependencies": {
-    "@types/node": "^9.6.8",
-    "@types/react": "^16.3.12",
+    "@types/node": "^10.3.5",
+    "@types/react": "^16.4.1",
+    "babel-core": "^6.26.3",
```
vlad-zhukov: Required for `webpack.config.babel.js`. Apparently the new version of `webpack-cli` doesn't install it anymore (or was it another pkg? no idea).

PR #291, opened 2018-08-29
Loaders updates (especially css-loader)

marcofugaro:
I updated all the most common dependencies to their latest version. Most of them have only minor updates, and so no breaking changes.

The two that had major updates are the file-loader ([release]()) and the url-loader ([release]()).

For the file-loader, I updated it anyway since webpack-blocks supports node6 to up as well.

For the css-loader, I had to handle the deprecation of the `minimize` option, I documented in the readme how to minify the css, and I updated the tests as best as I can. 

Let me know if I can handle the examples/tests in any better way. Also should I update the changelogs?


andywer:
Great stuff, @marcofugaro! 🙌

I am just thinking that maybe we should now provide a postcss/cssnano configuration for the production build of the sample app, since the css-loader's `minify` is gone. Not necessarily blocking this PR, though 😉

marcofugaro:
@vlad-zhukov alright done, updated also the postcss block since there were no breaking changes with the new major version

marcofugaro:
@andywer thanks! was wondering wether I should update also the `sampleapp`, but noticed you didn't use any external blocks, only the core. 

I can set up the cssnano minification in this PR or another if you want, let me know!

PR #257, opened 2018-03-02
@webpack-blocks/babel: Support for Babel 7.x

dmitmel:
[Babel 7.x](https://babeljs.io/blog/2017/09/12/planning-for-7.0) will be [released soon](https://babeljs.io/blog/2017/12/27/nearing-the-7.0-release), so it will be great if the Babel block fully supports the next Babel version.

PR #295, opened 2018-08-31
Support for babel-loader@8 / babel@7

boxfoot:
Add support for Babel 7 (for 2.0 webpack-blocks release).  Supercedes PR #257 which was out-of-date.

### DIFF
--- a/packages/babel/package.json
+++ b/packages/babel/package.json
@@ -15,13 +15,13 @@
   "repository": "andywer/webpack-blocks",
   "bugs": "https://github.com/andywer/webpack-blocks/issues",
   "dependencies": {
-    "babel-loader": "^7.1.4"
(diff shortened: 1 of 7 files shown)


packages/babel/package.json
```diff
@@ -1,5 +1,5 @@
 {
-  "name": "@webpack-blocks/babel",
+  "name": "@boxfoot/webpack-blocks-babel",
```
andywer: I spy with my little eye... 😉

boxfoot: Oops!  Fixed.  Sorry for the slow turnaround!

boxfoot:
@vlad-zhukov fixed.

andywer:
Looks like we need to update the babel preset in the sample app in order to fix the build. Looks fine otherwise 👍

PR #298, opened 2018-09-28
Remove css-loader and style-loader from the sass block

marcofugaro:
Followup of discussion in #293

### DIFF
--- a/packages/sass/README.md
+++ b/packages/sass/README.md
@@ -22,8 +22,7 @@ module.exports = createConfig([
 ```
 <!-- prettier-ignore-end -->
 
-**NOTE**: the `css()` block by default works on `.css` files, so you have to make it accept `.scss`
-files using the `match()` block.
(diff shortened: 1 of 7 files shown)


packages/sass/README.md
```diff
@@ -7,32 +7,24 @@ This is the `sass` block providing Sass support for webpack. Uses `node-sass` vi
...
 module.exports = createConfig([
   match(['*.scss', '!*node_modules*'], [
+    css(),
     sass(/* node-sass options */)
   ])
 ])
 ```
 <!-- prettier-ignore-end -->
 
+**NOTE**: the `css()` block by default works on `.css` files, so you have to make it accept `.scss`
```
andywer: I‘d like to change the wording here:
"Note: Use `match()` here to apply the `css()` block to `.scss` files"?

marcofugaro: Sure!

marcofugaro:
@andywer I updated the wording as you asked, maybe you missed it 🙂

PR #300, opened 2018-09-30
Upgrade all deps

vlad-zhukov:


packages/webpack/__e2e-fixtures__/babel-postcss-extract-text/webpack.config.js
```diff
@@ -14,7 +14,7 @@ const postcss = require('@webpack-blocks/postcss')
 const extractText = require('@webpack-blocks/extract-text')
 const { css } = require('@webpack-blocks/assets')
 const path = require('path')
-const precss = require('precss')
+const postcssVars = require('postcss-simple-vars')
```
vlad-zhukov: `precss` is outdated and brings lots of dependencies with itself, but its features are not required for these tests.

vlad-zhukov:
Ava fails with a weird babel error: 
```Error: You gave us a visitor for the node type TSUnknownKeyword but it's not a valid type```

I can also reproduce it locally on `release-2.0` branch, but CI doesn't...

**Comment by vlad-zhukov** [comment]
I guess the `ava` upgrade from my last PR didn't go well, downgrading the version works however.

**Comment by andywer** [comment]
PS: Downgrading AVA seems fine.

**Comment by vlad-zhukov** [comment]
To sum up: if you update from one of previous commits to [`abb2d05` Upgrade root deps (#299)](https://github.com/andywer/webpack-blocks/pull/266/commits/abb2d0557cc4e6264f4765207badba84c7172bd6) and run `yarn`, the `yarn ava` will run just fine. But on a fresh `node_modules` install this commad will fail due to a weird error in one of Babel plugin `ava` uses.

I've downgraded `ava` from `1.0.0-beta.8` to `1.0.0-beta.7`, update almost all other deps and made sure the fresh install passes all tests. Everything should be fine now! 🙏

**Comment by andywer** [comment]
Sounds good! Build is failing, though 🙃

We either need to find that dependency that's not compatible to node 6 or bump to minimum node 8. I would prefer continued support for node 6 if it's not too much effort.

**Comment by vlad-zhukov** [comment]
Damn, (`jsdom` v12 dropped support for `node` 6)[https://github.com/jsdom/jsdom/blob/master/Changelog.md#1200], downgraded it back to v11.

**Comment by vlad-zhukov** [comment]
@andywer yeah, we should stick to what `node` version `webpack` supports, [it's currently >=6.11.5](https://github.com/webpack/webpack/blob/86f56bf635659e9ab3d373f6e81dc66ee2391149/package.json#L86)

File: packages/sample-app/tsconfig.json [hunk 2]
```diff
@@ -5,7 +5,7 @@
     "removeComments": true,
     "preserveConstEnums": true,
     "sourceMap": true,
-    "target": "es5",
+    "target": "es2015",
```
vlad-zhukov: The newest `@types/node` require it. Node 6 supports ES6, so it's fine.

